PR #36, opened 2020-01-29
Lint package configuration 

### LINKED ISSUE #24: Adopt lint package for linting
>Google publicly shares their internal rules as package:pedantic in open-source. It represents what Google is enforcing internally throughout all Dart code. For a lint rule to be added to pedantic, Google has to change all code which doesn't follow the style.
>
>This strict practice results in only 27/150+ rules to be enabled. While some are contradictory to each other and can not enabled together, a big chunk of rules isn't enabled because it requires too much work to update all of Googles existing code.
>
>For developers outside Google, it is the norm to have separate lint rules per project. One project might enable more rules then others.
>
> lint enables a majority of lint rules, leaving out contradictory and very opinionated rules.

https://pub.dev/packages/lint

https://resocoder.com/2020/01/17/dart-linting-tutorial-healthy-flutter-codebase-with-analysis_options-yaml/

- [ ] make sure that lint warnings fail build on bitrise
- [ ] as we use monrepo check if we can use same lint for all projects

vova-beloded:
Fixes #24

- [ ] Add common analysis_options.yaml file

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,0 +1,12 @@
+include: package:lint/analysis_options.yaml
+
+# https://github.com/passsy/dart-lint
+linter:
+  rules:
+    - unawaited_futures # https://dart-lang.github.io/linter/lints/unawaited_futures.html
+
+# https://github.com/passsy/dart-lint/blob/master/lib/analysis_options.yaml
+analyzer:
+  errors:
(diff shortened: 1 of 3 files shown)


bootstraponline:
I added `Fixes #24` so that the issue is closed when the PR is merged.

PR #60, opened 2020-02-06
Build results widget implementation

### LINKED ISSUE #9: Create build results widget
* Design widget that reads in data from the model and renders according to the spec
* Write a component spec to verify the widget works as expected
* Write a FlutterDriver test to verify the widget works in a real browser
* Note that the individual bars in the graph should be clickable. Once clicked, they will open a link to the build on Jenkins/Buildkite/etc.

**FAQ**
1. How many bars should we render?
Let's start with 14 based on the design, but make it configurable.

2. Is the graph interactive?
Yes, the user should be able to click onto an individual bar and that will open a new URL. See https://buildkite.com/bazel

3. Should we use real data?
For now, mock out all the data.

Graph should resize based on screen size. See Google example:
https://google.github.io/charts/flutter/example/bar_charts/spark_bar

vova-beloded:
Fixes #9

## Checklist

- [x] Documented
- [x] Unit tested
- [x] Component tested
- [x] Driver tested


metrics/test/features/dashboard/presentation/widgets/bar_graph_test.dart
```diff
@@ -0,0 +1,221 @@
...
+      for (final barWidget in barWidgets) {
+        await tester.tap(find.byWidget(barWidget));
+
+        expect(tappedBarValue, barWidget.value);
+      }
+    },
+  );
+
+  testWidgets(
+    'Builds all bar data in a right order',
```
illia-romanenko: What if we clarify what "right" order means here?

metrics/test/features/dashboard/presentation/widgets/bar_graph_test.dart
```diff
@@ -0,0 +1,221 @@
...
+          matching: find.byType(GraphTestBar),
+        ));
+
+        expect(bar.value, BarGraphTestbed.graphBarTestData[i].value);
+      }
+    },
+  );
+
+  testWidgets(
+    'Builds the graph bars with the right height ratio',
```
illia-romanenko: Might be a good idea to clarify "right" here as well?

metrics/test/features/dashboard/presentation/widgets/build_result_bar_graph_test.dart
```diff
@@ -0,0 +1,140 @@
...
+    "Can't create widget without data",
+    (WidgetTester tester) async {
+      await tester.pumpWidget(const BuildResultBarGraphTestbed(data: null));
+
+      expect(tester.takeException(), isA<AssertionError>());
+    },
+  );
+
+  testWidgets(
+    "Creates the correct number of bars",
```
illia-romanenko: "Correct" seem be to be not clear word as well for it?

metrics/test/features/dashboard/presentation/widgets/build_result_bar_graph_test.dart
```diff
@@ -0,0 +1,140 @@
...
+      await tester.pumpWidget(const BuildResultBarGraphTestbed());
+
+      final barWidgets = tester.widgetList(find.byType(ColoredBar));
+
+      expect(barWidgets.length, buildResults.length);
+    },
+  );
+
+  testWidgets(
+    "Creates bars with ptoper colors",
```
illia-romanenko:
```suggestion
    "Creates bars with colors from [ColorConfig]",
```

PR #61, opened 2020-02-10
add third party api testing document

### LINKED ISSUE #44: Review 3rd party API testing techniques
As we integrate with 3rd party services like Jira & Slack we want to make sure that our integration is reliable.

As a result of this ticket completion, we want to have documented 3rd party API testing approach we plan to use for Jira & Slack integration testing.

vladlaba:
Fixes #44

### DIFF
--- a/docs/00_template.md
+++ b/docs/00_template.md
@@ -21,6 +21,7 @@
 # Design
 
 > Explain and diagram the technical design
+>
 > Identify risks and edge cases
 
 # API
--- a/docs/03_third_party_api_testing.md
+++ b/docs/03_third_party_api_testing.md
@@ -1,0 +1,96 @@
+# Testing Third Party API
+> Summary of the proposed change
+
+Research 3rd party API testing strategies.
+
+# References
+> Link to supporting documentation, GitHub tickets, etc.
+
+- [Testing Third Party Interactions](https://thoughtbot.com/blog/testing-third-party-interactions)
+
+# Motivation
+> What problem is this project solving?
+
+Rapidly detect regressions in API clients.
+
+# Goals
+> Identify success metrics and measurable goals.
+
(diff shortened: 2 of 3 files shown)


vladlaba:
@illia-romanenko @bootstraponline 

What do you think about adding some examples for each approach? 

The reason is that links contain examples in Ruby but as we're working with Dart it would be great to have some base examples in testing third party API without the need to follow one of provided link and reading the article.

illia-romanenko:
Thanks for the suggestion!

I think that we are fine with the current documentation. If we'll need to go further and use other methods we can think about it.

vladlaba:
> I have minor suggestion - swap Dart packages and Definitions - to provide more detailed description of the problem and then go to dart implementation details.

Thanks! I've placed Dart packages after techniques - looks better!

illia-romanenko:
@vladlaba we'll need to work on this a bit more utilizing template for documentation, let's discuss it tomorrow so I can give more details

bootstraponline:
|             | Adapter level | HTTP level  |
| ------------- | ------------- | ---- |
| **Stubbed** | Stub methods on adapter  | Stub HTTP requests |
| **Real**  | Real request to fake adapter | Real HTTP request to fake service |

Let's add a markdown table instead of using the image. This is more accessible.

bootstraponline:
Awesome work so far on the document! I have cleaned it up a bit and committed the changes.


# API
> What will the proposed API look like?

For the API section, I would like to document what the API will look like for the one solution we are proposing (mock web server).

bootstraponline:
Also for this pull request, it will be a `Squash and merge` because there is `one logical change` (the markdown document).

bootstraponline:
The docs look great! I updated the language a bit and approved.

PR #68, opened 2020-02-15
Add shell_words

### LINKED ISSUE #57: Port Buildkite Shellwords parser from Go to Dart
The task is to create a standalone Dart port of the shellwords library from Buildkite. Shellwords is useful for parsing strings in the same way as the UNIX Bourne shell. 

> A golang library for splitting command-line strings into words like a Posix or Windows shell would.

https://github.com/buildkite/shellwords

# Tasks
- [ ] Configure the shellwords job on Bitrise. https://app.bitrise.io/app/65e2c5934061cc08
- [x] Port parser https://github.com/buildkite/shellwords/blob/master/parser.go
- [x] Use the right quotes depending on OS https://github.com/buildkite/shellwords/blob/master/quote.go#L8
- [x] Port tests

bootstraponline:
Fixes #57

Dart port of https://github.com/buildkite/shellwords

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/shell_words/pubspec.yaml
+++ b/shell_words/pubspec.yaml
@@ -1,0 +1,14 @@
+name: shell_words
+description: A sample command-line application.
+# version: 1.0.0
+# homepage: https://www.example.com
+
+environment:
+  sdk: '>=2.7.0 <3.0.0'
+
+dependencies:
+  unicode: ^0.2.2
+
+dev_dependencies:
+  pedantic: ^1.8.0
+  test: ^1.6.0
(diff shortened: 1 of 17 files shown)


shell_words/pubspec.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+environment:
+  sdk: '>=2.7.0 <3.0.0'
+
+dependencies:
+  unicode: ^0.2.2
+
+dev_dependencies:
+  pedantic: ^1.8.0
+  test: ^1.6.0
```
madz: Should we upgrade all dependencies to the latest?

pedantic: ^1.8.0 to pedantic: ^1.9.0
pedantic 1.9.0 Enforce 17 new lint rules.

test package has latest version of 1.12.0. I think it has some substantial bug fixes and the only minimum SDK version is 2.4.0 which we are already in.

PR #70, opened 2020-02-17
Make widgets look more like designs

### LINKED ISSUE #67: Update widgets to look more like the designs
The widgets should look like the dark theme. The dark theme should be the default.

For the light theme, the widgets should look identical except the colors will be different.

## **Current theme**
image

## **Dark theme**
image

## **Blue / Light theme**
image

vova-beloded:
Fixes #67

### DIFF
--- a/metrics/lib/features/dashboard/presentation/strings/dashboard_strings.dart
+++ b/metrics/lib/features/dashboard/presentation/strings/dashboard_strings.dart
@@ -1,3 +1,7 @@
+/// Holds the strings for the dashboard screen.
+///
+/// Preferably, add the string to this file instead of hardcoding them into UI
+/// to make them available in tests and avoid code duplication.
 class DashboardStrings {
   static const String buildTaskName = 'Build task name';
   static const String performance = 'PERFORMANCE';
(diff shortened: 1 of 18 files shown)


bootstraponline:
@vova-beloded I broke the tests. Can we update the tests to dynamically reference the strings? That will solve the problem of having hard coded text duplicated in two places.

vova-beloded:
@bootstraponline, done. 
I've moved all the strings to the separate file to make them available in driver tests too. If we will import the UI components (dart:ui package) into the driver tests file, the test fails.

bootstraponline:
Awesome! Thanks for fixing this.

bootstraponline:
This looks good to merge. I'll let @illia-romanenko code review as well.

metrics/lib/features/dashboard/presentation/strings/dashboard_strings.dart
```diff
@@ -0,0 +1,11 @@
+class DashboardStrings {
```
illia-romanenko: Great job so far! I think it would be great to add documentation here, for class & method?

PR #83, opened 2020-02-20
Metric widgets improvements

### LINKED ISSUE #82: Improve bar chart widget display when there is less data then should be
Make sure that the bar chart widget has the ability to display differently when there is less data than expected (show empty slots). It should look like this 
image

Note empty slots displayed if data is not present.
If there is less data than expected there should be an ability to align data to the right (as this is what we want to do on the dashboard).

vova-beloded:
Fixes #82 

## Checklist

- [x] Documented
- [x] Component tested

### DIFF
--- a/metrics/lib/features/dashboard/presentation/widgets/placeholder_bar.dart
+++ b/metrics/lib/features/dashboard/presentation/widgets/placeholder_bar.dart
@@ -1,10 +1,15 @@
 import 'package:flutter/material.dart';
+import 'package:metrics/features/dashboard/presentation/widgets/bar_graph.dart';
 import 'package:metrics/features/dashboard/presentation/widgets/colored_bar.dart';
 
 /// Represents the placeholder bar of the [BarGraph].
 class PlaceholderBar extends StatelessWidget {
   final double width;
 
+  /// Creates the [PlaceholderBar].
+  ///
+  /// This bar displays the missing/empty build in [BarGraph].
+  /// [width] is the width of this bar.
   const PlaceholderBar({
     Key key,
     @required this.width,
--- a/metrics/lib/features/dashboard/presentation/widgets/placeholder_bar.dart
+++ b/metrics/lib/features/dashboard/presentation/widgets/placeholder_bar.dart
@@ -1,0 +1,29 @@
+import 'package:flutter/material.dart';
+import 'package:metrics/features/dashboard/presentation/widgets/colored_bar.dart';
+
+/// Represents the placeholder bar of the [BarGraph].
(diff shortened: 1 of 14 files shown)


metrics/lib/features/dashboard/presentation/widgets/placeholder_bar.dart
```diff
@@ -0,0 +1,29 @@
+import 'package:flutter/material.dart';
+import 'package:metrics/features/dashboard/presentation/widgets/colored_bar.dart';
+
+/// Represents the placeholder bar of the [BarGraph].
+class PlaceholderBar extends StatelessWidget {
+  final double width;
+
+  const PlaceholderBar({
```
illia-romanenko: Can we add documentation for the constructor?

metrics/lib/features/dashboard/presentation/widgets/sparkline_graph.dart
```diff
@@ -150,13 +152,22 @@ class SparklineGraph extends StatelessWidget {
...
 
     final sortedData = data.toList()..sort();
-    return DoubleSpan(sortedData.first.toDouble(), sortedData.last.toDouble());
+
+    final min = sortedData.first.toDouble();
+    double max = sortedData.last.toDouble();
+
+    final axisLength = max - min;
+    if (axisLength < _minChartAxisLength) {
+      max += _minChartAxisLength - axisLength;
```
illia-romanenko: Shouldn't we just add _minChartAxisLength?

metrics/test/features/dashboard/presentation/widgets/sparkline_graph_test.dart
```diff
@@ -256,6 +256,37 @@ void main() {
       expect(chartLine.fill.gradient.colors.first, accentColor);
     },
   );
+
+  testWidgets(
+    "Draws the graph even if one of the graph axes is too short",
```
illia-romanenko:
```suggestion
    "Draws the graph even if one of the graph data axes is 0",
```

metrics/lib/features/dashboard/presentation/widgets/sparkline_graph.dart
```diff
@@ -158,10 +158,13 @@ class SparklineGraph extends StatelessWidget {
   DoubleSpan _createSpan(List<num> data) {
     if (data == null || data.isEmpty) return null;
 
-    final sortedData = data.toList()..sort();
+    double min;
+    double max;
```
illia-romanenko: Can we initialize it with first to make sure that we are not doing null check in the loop later?

PR #124, opened 2020-03-27
Implement Jenkins client adapter

### LINKED ISSUE #128: Jenkins Ci Adapter
Create Jenkins Ci Adapter that adopts Jenkins Client - to Ci Client interface.

vladlaba:
Fixes #128 

## Checklist

- [x] Documented


metrics/ci_integrations/lib/jenkins/adapter/jenkins_ci_client_adapter.dart
```diff
@@ -0,0 +1,182 @@
...
+    JenkinsQueryLimits limits = const JenkinsQueryLimits.empty(),
+  }) async {
+    final newBuildsFetchResult =
+        await jenkinsClient.fetchBuilds(projectId, limits: limits);
+    _throwIfInteractionUnsuccessful(newBuildsFetchResult);
+    return newBuildsFetchResult.result;
+  }
+
+  /// Checks the given [interactionResult] to be [InteractionResult.isSuccess].
+  ///
```
illia-romanenko:
```suggestion
```
nit: we probably don't need that initial as what comes next describes it well enough.

metrics/ci_integrations/lib/jenkins/adapter/jenkins_ci_client_adapter.dart
```diff
@@ -0,0 +1,182 @@
...
+          await jenkinsClient.fetchArtifactByRelativePath(
+        jenkinsBuild.url,
+        coverageArtifact.relativePath,
+      );
+
+      _throwIfInteractionUnsuccessful(artifactContentFetchResult);
+
+      final artifactContent = artifactContentFetchResult.result;
+      coverage = CoverageJsonSummary.fromJson(artifactContent);
+    }
```
illia-romanenko: I suggest we extract this code to a separate method and pass coverage to this _mapJenkinsBuild as a parameter due to SRP.

metrics/ci_integrations/lib/jenkins/adapter/jenkins_ci_client_adapter.dart
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  /// Processes the given [builds] to the list of [BuildData]s.
+  ///
+  /// The [jobName] is used to identify a building job for the builds.
+  /// The [startFromBuildNumber] is used to filter builds which number is less
+  /// than or equal to this value. This allows to avoid fetching old builds
+  /// since the range specifier in Jenkins API only provides ability to set
+  /// fetch limits but not to filter data to fetch.
```
illia-romanenko: I'm not sure how we avoid fetching here as this piece called after fetching? Could you please explain this part a bit more?

metrics/ci_integrations/lib/jenkins/adapter/jenkins_ci_client_adapter.dart
```diff
@@ -0,0 +1,182 @@
...
+      limits: JenkinsQueryLimits.endBefore(0),
+    );
+    final lastBuild = buildingJob.lastBuild;
+    int numberOfBuilds = lastBuild.number - build.buildNumber;
+
+    if (numberOfBuilds <= 0) return [];
+
+    List<JenkinsBuild> builds;
+    while (builds == null) {
+      final buildingJob = await _fetchBuilds(
```
illia-romanenko:
```suggestion
      final buildingJobForNumberOfBuilds = await _fetchBuilds(
```
How about changing it like this, as we already have buildingJob defined earlier in this method so it might make sense to change the name or we can `Extract Method` here to make this while more granular?

metrics/ci_integrations/test/common/config/ci_config_test.dart
```diff
@@ -0,0 +1,61 @@
...
+      () {
+        expect(
+          () => CiConfig(
+            ciProjectId: ciProjectId,
+            storageProjectId: null,
+          ),
+          throwsArgumentError,
+        );
+      },
+    );
```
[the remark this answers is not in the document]
vladlaba: This tests that we cannot create the `CiConfig` instance with no `storageProjectId` given and the next one - with no `ciProjectId`. I'm not sure we should get rid of this since it is a different test case.

metrics/ci_integrations/lib/jenkins/adapter/jenkins_ci_client_adapter.dart
vladlaba: I followed the Effective Dart documentation guide here: https://dart.dev/guides/language/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph. I think we should follow these rule.

metrics/ci_integrations/lib/jenkins/adapter/jenkins_ci_client_adapter.dart
vladlaba: The Jenkins range-specifier works as a `sublist` for the resulting array-like property it is applied to. And not filters data. Therefore, the following situation is possible: 
- The last build in storage has the `buildNumber` equal to 10.
- There are builds with numbers 11, 12, 13, 14, 15 respectively. 
- A user deletes the builds with numbers 12 and 13. 
- The algorithm fetches the last build from CI - the result is 15. 
- Since the last build in CI is 15 and the last one in storage is 10, we should fetch 5 builds. 
- The algorithm fetches 5 builds - the result is 9, 10, 11, 14, 15. But we don't need to store 9 and 10, so we should filter.

So the range-specifier does not allow to handle such a situation and we have to add additional parameters.

PR #144, opened 2020-04-06
Unique firestore document id

### LINKED ISSUE #132: Add unique Firestore document id for CI integrations
Create a unique key approach: "project id"_"build id" so if multiple integrations started at the same time - data will be consistent.

dmytroSolidSoftware:
Fixes #132 

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/metrics/ci_integrations/lib/ci_integration/ci_integration.dart
+++ b/metrics/ci_integrations/lib/ci_integration/ci_integration.dart
@@ -43,7 +43,12 @@ class CiIntegration {
       }
 
       if (newBuilds != null && newBuilds.isNotEmpty) {
-        await storageClient.addBuilds(storageProjectId, newBuilds);
+        await storageClient.addBuilds(
+          storageProjectId,
+          newBuilds
+              .map((build) => build.copyWith(projectId: storageProjectId))
+              .toList(),
+        );
         return const InteractionResult.success(
           message: 'The project has been updated successfully!',
         );
--- a/metrics/ci_integrations/lib/ci_integration/ci_integration.dart
+++ b/metrics/ci_integrations/lib/ci_integration/ci_integration.dart
@@ -43,11 +43,12 @@ class CiIntegration {
       }
 
       if (newBuilds != null && newBuilds.isNotEmpty) {
+        final mappedBuilds = newBuilds
+            .map((build) => build.copyWith(projectId: storageProjectId))
+            .toList();
         await storageClient.addBuilds(
           storageProjectId,
-          newBuilds
-              .map((build) => build.copyWith(projectId: storageProjectId))
-              .toList(),
+          mappedBuilds,
         );
         return const InteractionResult.success(
           message: 'The project has been updated successfully!',
--- a/metrics/ci_integrations/lib/ci_integration/ci_integration.dart
+++ b/metrics/ci_integrations/lib/ci_integration/ci_integration.dart
@@ -43,12 +43,9 @@ class CiIntegration {
       }
 
       if (newBuilds != null && newBuilds.isNotEmpty) {
-        final mappedBuilds = newBuilds
-            .map((build) => build.copyWith(projectId: storageProjectId))
-            .toList();
         await storageClient.addBuilds(
           storageProjectId,
-          mappedBuilds,
+          newBuilds,
         );
         return const InteractionResult.success(
           message: 'The project has been updated successfully!',
--- a/metrics/ci_integrations/lib/firestore/adapter/firestore_storage_client_adapter.dart
+++ b/metrics/ci_integrations/lib/firestore/adapter/firestore_storage_client_adapter.dart
@@ -26,12 +26,9 @@ class FirestoreStorageClientAdapter implements StorageClient {
       final futures = <Future>[];
 
       for (final build in builds) {
-        futures.add(collection.add(
-          build.toJson()
-            ..addAll(
-              {'projectId': project.id},
-            ),
-        ));
+        final map = build.toJson();
+        map['projectId'] = project.id;
+        futures.add(collection.document(build.documentId).create(map));
       }
       await Future.wait(futures);
     } on GrpcError catch (e) {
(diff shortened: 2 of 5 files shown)


metrics/core/lib/src/data/model/build_data.dart
```diff
@@ -25,12 +27,40 @@ class BuildData extends Build implements DataModel {
           coverage: coverage,
         );
 
+  /// Unique id for multiple integrations
```
vladlaba, line 30:
```suggestion
  /// The unique identifier of this build. 
```
or 
```suggestion
  /// The unique identifier of a document representing this build.
```
What do you think?

metrics/core/lib/src/data/model/build_data.dart
```diff
@@ -25,12 +27,40 @@ class BuildData extends Build implements DataModel {
...
+  String get documentId => '${projectId}_$id';
+
+  BuildData copyWith({
```
vladlaba, line 33: Should we document this method?

metrics/core/lib/src/domain/entities/build.dart
```diff
@@ -16,6 +17,7 @@ class Build extends Equatable {
   /// Creates the [Build].
   ///
   /// [id] is the unique identifier of this build.
+  /// [projectId] is the unique identifier of the project related with this build.
```
vladlaba, line 20:
```suggestion
  /// [projectId] is the unique identifier of the project this build belongs to.
```
or 
```suggestion
  /// [projectId] is the unique identifier of the project this build is related to.
```
?

metrics/ci_integrations/lib/ci_integration/ci_integration.dart
```diff
@@ -43,7 +43,12 @@ class CiIntegration {
       }
 
       if (newBuilds != null && newBuilds.isNotEmpty) {
-        await storageClient.addBuilds(storageProjectId, newBuilds);
+        await storageClient.addBuilds(
+          storageProjectId,
+          newBuilds
+              .map((build) => build.copyWith(projectId: storageProjectId))
+              .toList(),
```
vladlaba, line 50: Maybe move this to a variable and pass it to the `addBuilds` method? I suppose this should improve readability - what do you think?

metrics/core/lib/src/data/model/build_data.dart
```diff
@@ -25,12 +27,44 @@ class BuildData extends Build implements DataModel {
           coverage: coverage,
         );
 
+  /// The unique identifier of this build.
+  String get documentId => '${projectId}_$id';
```
vladlaba, line 31:
```suggestion
  String get documentId => '${projectId}_$buildNumber';
```
It seems like we decided to use the build number as the second part of this value, didn't we? 

cc @illia-romanenko

metrics/core/lib/src/data/model/build_data.dart
```diff
@@ -25,12 +27,44 @@ class BuildData extends Build implements DataModel {
...
+
+  /// Creates the new instance of the [BuildData] based on current instance.
+  ///
+  /// If any of the passed parameters are null, or parameter isn't specified,
+  /// the value will be copied from the current instance.
```
vladlaba, line 36:
```suggestion
  /// Creates a new instance of the [BuildData] based on this one.
  ///
  /// If an argument is `null` (either `null` or nothing is passed) the value is copied 
  /// from the corresponding field of this instance.
```
?

metrics/ci_integrations/lib/firestore/adapter/firestore_storage_client_adapter.dart
```diff
@@ -26,12 +26,9 @@ class FirestoreStorageClientAdapter implements StorageClient {
...
 
       for (final build in builds) {
-        futures.add(collection.add(
-          build.toJson()
-            ..addAll(
-              {'projectId': project.id},
-            ),
-        ));
+        final map = build.toJson();
+        map['projectId'] = project.id;
```
illia-romanenko, line 30: I think we should build.copyWith(projectId: project.id) now as we don't want to keep toJson details (like field name) here.

metrics/core/lib/src/data/model/build_data.dart
illia-romanenko: Can we summarize our discussion about adding separate getter for documentId? vs setting id with projectId + buildNumber?

metrics/core/lib/src/data/model/build_data.dart
illia-romanenko: `this one` feels a bit javascriptey - maybe based on `current instance` would be better?
here is an example from `TextTheme`

> Creates a copy of this text theme but with the given fields replaced with the new values.

Maybe we can use something similar?

vladlaba: The reason is that Effective Dart [recommends](https://dart.dev/guides/language/effective-dart/documentation#prefer-using-this-instead-of-the-to-refer-to-a-members-instance) using `this` when referencing the instance members. Maybe we can change this to `this instance`. Generally, I suppose that using `current` may be confusing making the BuildData a trackable entity that changes over time (like a user).

metrics/core/lib/src/data/model/build_data.dart
vladlaba: Yes, we can get rid of this getter and use the `id` field. This makes sense since we are using [BuildData.id] as a document ID in Metrics. Thanks, I missed that in review!

metrics/core/lib/src/data/model/build_data.dart
illia-romanenko: While they say about `this` they are not saying about `this one` - I feel like example description from TextTheme (that includes `this`) is generally better.

metrics/core/lib/src/data/model/build_data.dart
```diff
@@ -25,12 +27,42 @@ class BuildData extends Build implements DataModel {
           coverage: coverage,
         );
 
+  /// Creates a copy of this [BuildData] but with the given fields
+  /// replaced with the new values.
+  ///
+  /// If any of the passed parameters are null, or parameter isn't specified,
+  /// the value will be copied from the current instance.
```
illia-romanenko, line 34:
```suggestion
```
Maybe just simplify?

PR #152, opened 2020-04-07
Dart code style (test description style)

### LINKED ISSUE #105: Test description style
As we see that tests descriptions style varies - we need to prepare a document with the test description style guide so our documentation is consistent.

illia-romanenko:
Fixes #105 

## Checklist

- [x] Documented

### DIFF
--- a/docs/10_dart_code_style.md
+++ b/docs/10_dart_code_style.md
@@ -1,7 +1,7 @@
 # Dart code style
 > Summary of the proposed change
 
-Make sure that our code style is consistent for parts that are not covered in Dart code style.
+Make sure our code style is consistent in parts that are not covered in the Dart code style.
 
 # References
 > Link to supporting documentation, GitHub tickets, etc.
@@ -13,7 +13,7 @@ all code style, documentation, design and usage conventions.
 # Motivation
 > What problem is this project solving?
 
-As Dart has a great code styleavailable
+As Dart has a great code style available
 that we use on this project we noticed that in some real-world cases there is no guidance,
 so here we collect all cases that we think important to clarify to make sure that the code style
 is consistent through the project.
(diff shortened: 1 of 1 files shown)


docs/10_dart_code_style.md
```diff
@@ -0,0 +1,100 @@
...
+    ```dart
+    test("should throw ArgumentError trying to create an instance with null CI client", 
+    ```
+3. Test description not inside Test Group starts with Capital letter:
+    ```dart
+    testWidgets("Can't create widget without data",
+   ```
+4. Group and test descriptions should not end with a dot (.).
+5. If method name (or named constructor) is used in test description it should start with dot (.)
+and should have parenthesis:
```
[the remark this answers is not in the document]
illia-romanenko: Thanks, pushed updated code.

docs/10_dart_code_style.md
```diff
@@ -0,0 +1,100 @@
...
+    ```
+6. Use Test Group descriptions with class under test name:
+    ```dart
+    group("ReceiveProjectMetricUpdates", () {
+       .....
+       test("loads all fields in the performance metrics", () {
+    ```
+7. Multiline test descriptions shouldn't be used as such tests can't 
+be started in [Intellij](https://youtrack.jetbrains.com/issue/WEB-44842).
+8. Prefer using double quotes.
```
vladlaba, line 64: Maybe we should add examples for these points as for others above? We can use the same style as the Effective Dart do. I.e., provide `bad` and `good` examples and use `DO`, `DO NOT`, etc. words.

illia-romanenko: Thanks @vladlaba - please check the new version and let me know what you think 💭

illia-romanenko:
Thanks, guys!

PR #183, opened 2020-05-06
Add tests for Jenkins Source Client Adapter

### LINKED ISSUE #151: [cli] Test Jenkins adapter
Add Jenkins adapter tests (mockito can be used for this).

vladlaba:
Fixes #151 

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/metrics/ci_integrations/test/source/jenkins/adapter/jenkins_source_client_adapter_test.dart
+++ b/metrics/ci_integrations/test/source/jenkins/adapter/jenkins_source_client_adapter_test.dart
@@ -102,7 +102,7 @@ void main() {
       ".fetchBuilds() fetches builds which are not building",
       () {
         final jenkinsBuilds = [
-          createJenkinsBuild(buildNumber: 1),
+          createJenkinsBuild(buildNumber: 1, building: false),
           createJenkinsBuild(buildNumber: 2, building: true),
         ];
         final expected = [createBuildData(buildNumber: 1)];
@@ -211,7 +211,7 @@ void main() {
     );
 
     test(
-      ".fetchBuilds() results with correct list of fetched builds",
+      ".fetchBuilds() maps fetched builds statuses according to specification",
       () async {
         const jenkinsResults = [
           JenkinsBuildResult.failure,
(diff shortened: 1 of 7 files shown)


metrics/ci_integrations/test/source/jenkins/adapter/jenkins_source_client_adapter_test.dart
```diff
@@ -0,0 +1,505 @@
...
+        ).thenAnswer((_) => responses.error<Map<String, dynamic>>());
+
+        final result = adapter.processJenkinsBuilds(jenkinsBuilds, jobName);
+
+        expect(result, throwsStateError);
+      },
+    );
+
+    test(
+      ".processJenkinsBuilds() fetches a coverage for each build",
```
illia-romanenko, line 170: I wanted to start a broader discussion here about making it visible for testing - while we test this method in isolation and much simpler we can't guarantee that it's used in the fetchBuilds & fetchBuildsAfter - so it feels like that we need to test it through those interface methods even if that means more repetitive tests. What do you think?

metrics/ci_integrations/test/source/jenkins/adapter/jenkins_source_client_adapter_test.dart
```diff
@@ -0,0 +1,597 @@
...
+    setUp(() {
+      reset(jenkinsClientMock);
+      responses.reset();
+    });
+
+    test(
+      ".fetchBuilds() fetches builds which are not building",
+      () {
+        final jenkinsBuilds = [
+          createJenkinsBuild(buildNumber: 1),
```
illia-romanenko, line 105:
```suggestion
          createJenkinsBuild(buildNumber: 1, building: false),
```
As it seems relevant to the test and not rely on the default value, readability should be improved as well with this change.

metrics/ci_integrations/test/source/jenkins/adapter/jenkins_source_client_adapter_test.dart
```diff
@@ -0,0 +1,597 @@
...
+
+        expect(
+          result,
+          completion(hasLength(equals(expectedBuildsLength))),
+        );
+      },
+    );
+
+    test(
+      ".fetchBuilds() results with correct list of fetched builds",
```
illia-romanenko, line 214:
```suggestion
      ".fetchBuilds() maps fetched builds statuses according to specification",
```

metrics/ci_integrations/test/source/jenkins/adapter/jenkins_source_client_adapter_test.dart
```diff
@@ -0,0 +1,597 @@
...
+        )).thenAnswer(responses.fetchBuilds);
+
+        final result = adapter.fetchBuildsAfter(jobName, build);
+
+        expect(result, throwsStateError);
+      },
+    );
+
+    test(
+      ".fetchBuildsAfter() fetches new builds performed after the given build",
```
illia-romanenko, line 342: It looks like we test mapping here, feels like this test could be simplified to highlight "after" component

PR #190, opened 2020-05-08
Update testing documentation

### LINKED ISSUE #189: Update testing documentation to include latest clarifications
Don't use should in test description.
Empty Mock class shouldn't be documented.
Document all classes (Testbeds, Stubs, etc..) except empty Mocks, Fakes.

DO NOT document Mock classes. 
DO document Mock static fields.
DO NOT document Fake classes.
DO document Fake custom properties.
DO document Testbeds, Stubs, etc. used in testing.
AVOID document fields of test data classes.

illia-romanenko:
Fixes #189 

## Checklist

- [x] Documented

### DIFF
--- a/docs/10_dart_code_style.md
+++ b/docs/10_dart_code_style.md
@@ -141,22 +141,23 @@ be started in [Intellij](https://youtrack.jetbrains.com/issue/WEB-44842)
     ```
     
 11. DO document Mock static fields
-     Though it is assumed that Mock classes do not have any fields, sometimes it appears to be very pragmatic to add some _static_ fields to them in order to simplify code and tests structure, follow DRY or KISS, etc. (see Dart Mockito [best practices](https://pub.dev/packages/mockito#best-practices))
+    
+    Though it is assumed that Mock classes do not have any fields, sometimes it appears to be very pragmatic to add some _static_ fields to them in order to simplify code and tests structure, follow DRY or KISS, etc. (see Dart Mockito [best practices](https://pub.dev/packages/mockito#best-practices))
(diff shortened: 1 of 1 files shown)


docs/10_dart_code_style.md
```diff
@@ -92,28 +92,125 @@ be started in [Intellij](https://youtrack.jetbrains.com/issue/WEB-44842)
...
+    ```
+    
+    `Bad:`
+              
+    ```dart
+    /// This class mocks [SourceClient] to help with testing.
+    class SourceClientMock extends Mock implements SourceClient {}
+    ```
+    
+11. DO document Mock static fields
```
vladlaba, line 143:
```suggestion
11. DO document Mock static fields
     Though it is assumed that Mock classes do not have any fields, sometimes it appears to be very pragmatic to add some _static_ fields to them in order to simplify code and tests structure, follow DRY or KISS, etc. (see Dart Mockito [best practices](https://pub.dev/packages/mockito#best-practices))
     
    `Good:`

    ```dart
    class CoolClassMock extends Mock implements CoolClass {
    /// The really useful field that allows doing something extra-cool 
    /// in order to prevent something very bad.
    static const String reallyUsefulField = 'Really Useful Value';
    ```

    `Bad:`

    ```dart
    class CoolClassMock extends Mock implements CoolClass {
    static const String reallyUsefulField = 'Really Useful Value';
    ```
```

It seems like we should follow the same style for each rule we're adding. Hence, even if we don't have a real example in it, a developer should always be able to follow a pattern.

PR #230, opened 2020-06-09
Circle percentage metrics improvements

### LINKED ISSUE #229: Change CirclePercentage to defined architecture

vova-beloded:
Fixes #229 

## Checklist

- [x] Documented
- [x] Unit tested
- [x] Component tested
- [x] Driver tested


metrics/web/lib/common/presentation/graphs/circle_percentage.dart
```diff
@@ -131,7 +139,7 @@ class _CirclePercentageState extends State<CirclePercentage>
     );
   }
 
-  /// Animates the [AnimationController] to the [widget.value] if it is not null.
+  /// Animates the [AnimationController] to the [widget.stability] if it is not null.
```
illia-romanenko, line 142:
```suggestion
  /// Animates the [AnimationController] to the [widget.value] if it is not null.
```

metrics/web/lib/common/presentation/graphs/circle_percentage.dart
```diff
@@ -1,53 +1,65 @@
...
     this.valueStrokeWidth = 5.0,
     this.padding = EdgeInsets.zero,
-    this.valueColor,
-    this.strokeColor,
+    Color valueColor,
+    Color strokeColor,
     this.backgroundColor,
-  })  : assert(value == null || (value >= 0 && value <= 1)),
+  })  : assert(placeholder != null),
+        assert(value == null || (value >= 0 && value <= 1)),
```
illia-romanenko, line 60: Should we be more fail-proof here? As we use ViewModels now they don't guarantee that the value is within those bounds - so what if this widget can accept any double - but if it's not within bounds - the value is pushed to be within bounds, like -1 -> 0 and values more than 1 are treated as 1? Should we discuss it a bit more?

metrics/web/lib/base/presentation/graphs/circle_percentage.dart
```diff
@@ -37,28 +37,22 @@ class CirclePercentage extends StatefulWidget {
 
   /// Creates the circle percentage graph.
   ///
-  /// The [value] must be either `null` or from 0.0 (inclusive) to 1.0 (inclusive).
-  /// The [placeholder] must not be `null`. The [placeholder] is displayed,
-  /// if the [value] is `null`.
-  ///
```
illia-romanenko: It feels like that we need this description still (or at least some indication how it works - that it fits the value in some bounds, etc..)

PR #304, opened 2020-07-06
Add tests for project groups notifier

### LINKED ISSUE #303: Add tests for ProjectGroupsNotifier class

vova-beloded:
Fixes #303 

## Checklist

- [x] Documented
- [x] Unit tested
- [x] Component tested
- [x] Driver tested

### DIFF
--- a/metrics/web/test/project_groups/presentation/state/project_groups_notifier_test.dart
+++ b/metrics/web/test/project_groups/presentation/state/project_groups_notifier_test.dart
@@ -31,7 +31,7 @@ void main() {
     final projectGroup = ProjectGroup(
       id: projectGroupId,
       name: projectGroupName,
-      projectIds: const [],
+      projectIds: List.from([]),
     );
 
     const project = ProjectModel(id: 'projectId', name: 'projectName');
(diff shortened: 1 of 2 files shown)


metrics/web/test/project_groups/presentation/state/project_groups_notifier_test.dart
```diff
@@ -0,0 +1,874 @@
...
+    final receiveProjectGroupUpdates = ReceiveProjectGroupUpdatesMock();
+
+    final projectGroupsNotifier = ProjectGroupsNotifier(
+      receiveProjectGroupUpdates,
+      addProjectGroupUseCase,
+      updateProjectGroupUseCase,
+      deleteProjectGroupUseCase,
+    );
+
+    setUpAll(() async {
```
illia-romanenko, line 63: It looks like the count of selected projects is not updated on UI when editing group:
See that only 1 project select but the count is 6:

image

I think it worked before - or maybe there is some issue with my build - let me know.

metrics/web/test/project_groups/presentation/state/project_groups_notifier_test.dart
```diff
@@ -0,0 +1,874 @@
...
+import 'package:metrics/project_groups/presentation/view_models/project_checkbox_view_model.dart';
+import 'package:metrics/project_groups/presentation/view_models/project_group_card_view_model.dart';
+import 'package:metrics/project_groups/presentation/view_models/project_group_dialog_view_model.dart';
+import 'package:mockito/mockito.dart';
+import 'package:test/test.dart';
+
+import '../../../test_utils/matcher_util.dart';
+
+void main() {
+  group("ProjectGroupsNotifier", () {
```
illia-romanenko, line 27: Can we also test the case when a project group contains a project id of project that was removed? As we don't provide the UI for the projects configuration - I want to make sure that if somebody misconfigured projects - it will continue to function normally.

PR #312, opened 2020-07-09
Minor enhancements to metrics deployment documentation

kowalczykp:
Hey, I've tried deploying the metrics application on firebase, without prior experience with flutter and firebase, and I did not encounter any major issues while following the docs, so **nicely done on the docs and setup**!

There were just 2 small inconstancies which are being addressed via this PR.

Since this is a monorepo, it'd be great if nested dependencies would be handled more transparently somehow, as f.e. I've faced issue, where `firebase-functions`  dep was missing when deploying the seed function, due to multiple package.json existing in different directories, each requiring manual `npm install`

Declarative approach to setting up firebase(rather than a detailed step-by-step doc), might be worth trying out (f.e. via terraform) - I can try to PoC that, if you'd be interested.

## Checklist

- [x] Documented
- [ ] Unit tested
- [ ] Component tested
- [ ] Driver tested

### DIFF
--- a/docs/09_firebase_deployment.md
+++ b/docs/09_firebase_deployment.md
@@ -100,7 +100,7 @@ To configure the Flutter for Web application to use recently created Firestore D
 and go to the project setting (tap on the setting gear icon near the `Project Overview` on top of the left panel and select `Project settings`.
 2. Scroll down and find your Firebase Web Application. 
 3. Go to `Firebase SDK snippet` of your application, select `Config` and copy the generated code.
-4. Go to the `web/index.js` file in the application directory and replace the following piece of code with the copied one in step 3: 
+4. Go to the `web/index.html` file in the application directory and replace the following piece of code with the copied one in step 3:
     ```
         var firebaseConfig = {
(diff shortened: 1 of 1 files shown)


illia-romanenko:
Thanks for your feedback and changes to the documentation @kowalczykp, we glad that it worked fine!

As for multiple `package.json` - we can try to merge them into one to simplify the process, since functions are only for seed - we didn't feel right to join them right away also I feel like that there was some other reason for this split as well.

Overall we agree that it's better to automate this process but we didn't have much time for it. If @bootstraponline is OK with trying Terraform - I think that it could be a great improvement for deployments (this is where you certainly have the most experience - we don't have much of the experience with Terraform).

Thanks,
Illia

bootstraponline:
> Overall we agree that it's better to automate this process but we didn't have much time for it. If @bootstraponline is OK with trying Terraform - I think that it could be a great improvement for deployments (this is where you certainly have the most experience - we don't have much of the experience with Terraform).

Terraform sounds awesome. I'd love for @kowalczykp to contribute this.

illia-romanenko:
Great! Really excited to see PoC. Thanks, @bootstraponline !

PR #349, opened 2020-07-23
Create theme data for project groups dropdown

### LINKED ISSUE #348: Create theme data for project groups dropdown

vova-beloded:
Fixes #348 

## Checklist

- [x] Documented
- [x] Unit tested


metrics/web/lib/common/presentation/metrics_theme/config/color_config.dart
```diff
@@ -24,4 +24,7 @@ class ColorConfig {
   static const Color lightInputColor = Colors.white;
   static const Color darkInputColor = Color(0xFF0d0d0d);
   static const Color inputSecondaryTextColor = Color(0xFF545459);
+  static const Color darkDropdownBorderColor = Color(0xFF878799);
+  static const Color darkDropdownHoverColor = Color(0xFF1d1d20);
+  static const Color darkDropdownHoverBorderColor = Color(0xFF37373f);
```
illia-romanenko, line 29: Recently we are trying to move all theme-specific colors to the theme itself - so the suggestion is to move those colors to the Dark theme - and create similar colors for the Light theme. The reasoning is that we'll still have to do it in the future when we'll work on the light theme so in this case (or have 2 colors in the config light and dark - see last sentence why it's probably not a good idea), it would make much sense to move it to the specific theme. Also changing updating light theme should not affect Dark theme files (or there shouldn't be 2 reasons for the change of Color config).

metrics/web/lib/common/presentation/metrics_theme/model/project_groups_dropdown_theme_data.dart
```diff
@@ -0,0 +1,50 @@
...
+  final Color closedButtonBackgroundColor;
+
+  /// The background [Color] of the project group dropdown button when hovered.
+  final Color hoverBorderColor;
+
+  /// The border [Color] of the project group dropdown button when open.
+  final Color openedButtonBorderColor;
+
+  /// The border [Color] of the project group dropdown button when closed.
+  final Color closedButtonBorderColor;
```
illia-romanenko, line 30: What if we group parameters by state here:

openedButtonBorderColor
openedButtonBackgroundColor

closedButtonBorderColor
closedButtonBackgroundColor

hoverBorderColor
hoverBackgroundColor

Would it improve readability

metrics/web/lib/common/presentation/metrics_theme/model/project_groups_dropdown_theme_data.dart
```diff
@@ -0,0 +1,50 @@
+import 'package:flutter/material.dart';
+
+/// Stores the theme data for the project group dropdown.
+class ProjectGroupsDropdownThemeData {
+  /// A default [Color] of the dropdown button border.
+  static const Color _defaultBorderColor = Colors.transparent;
+
+  /// The background [Color] for main elements of the project group dropdown.
```
illia-romanenko, line 8: can we clarify what "main elements" are (maybe some clarification or examples?)

PR #388, opened 2020-08-03
Change google fonts to assets

### LINKED ISSUE #385: Change google fronts to assets due to font weight issues
Since the GoogleFonts package has an [issue](https://github.com/material-foundation/google-fonts-flutter/issues/35) with the font weights we should change our implementation and use the fonts from assets instead of GoogleFonts. Also, we should test such widgets for current applying the font weights: 

- PerformanceSparklineGraph

- ThemedCirclePercentage

solid-maxim:
Fixes #385


metrics/web/pubspec.yaml
```diff
@@ -75,12 +74,21 @@ flutter:
...
+        - asset: fonts/Roboto-Bold.ttf
+          weight: 700
+        - asset: fonts/Roboto-Light.ttf
+          weight: 300
+        - asset: fonts/Roboto-Medium.ttf
+          weight: 500
+        - asset: fonts/Roboto-Regular.ttf
+          weight: 400
+        - asset: fonts/Roboto-Thin.ttf
+          weight: 100
```
illia-romanenko, line 91: Could we please sort by weight to improve readability?

metrics/web/pubspec.yaml
```diff
@@ -75,12 +74,21 @@ flutter:
...
   #   - family: Trajan Pro
```
illia-romanenko, line 92: Should we remove those commented lines as we did for the previous part of it?

metrics/web/lib/main.dart
```diff
@@ -37,12 +36,10 @@ class _MyAppState extends State<MyApp> {
             ),
             themeMode: isDark ? ThemeMode.dark : ThemeMode.light,
             theme: ThemeData(
+              fontFamily: 'Roboto',
```
illia-romanenko, line 39: What if we move this font family to a constant?

PR #497, opened 2020-09-02
fix: Update chromedriver version

solid-maxim:


illia-romanenko (approved):
LGTM - thanks @solid-maxim @vova-beloded !

PR #503, opened 2020-09-03
fix: Remove driver tests from the metrics GitHub action

vova-beloded:


illia-romanenko (approved):
LGTM, thanks @vova-beloded

PR #504, opened 2020-09-03
fix: Enable driver tests and add post frame callback to the loading page

solid-maxim:


illia-romanenko (approved):
LGTM - thanks @vova-beloded @solid-maxim! Great job finding this issue.

PR #519, opened 2020-09-09
fix: Minor documentation and code improvements

solid-vladlaba:


illia-romanenko (approved):
LGTM - thanks for spotting those @vladlaba!

PR #527, opened 2020-09-11
fix: Update Metrics Logo icon in App Bar

solid-daniyilyevtyushkin:
Fixes #
Now Metrics Logo in App Bar has fixed proper size


solid-vladlaba (approved):
LGTM, thanks @solid-daniyilyevtyushkin!

PR #537, opened 2020-09-11
fix: Dropdown Item test descritions and readability fixes

solid-daniyilyevtyushkin:
Fixes #


solid-vladlaba (approved):
LGTM, thanks @solid-daniyilyevtyushkin !

PR #558, opened 2020-09-16
fix: Remove font-weight on the checked checkbox

### LINKED ISSUE #554: Font weight in checkbox item is changing after selecting checkbox
**Describe the bug**
Font (weight) shouldn't change on the checkbox selection.
image

**To Reproduce**
1. Go to project groups dialog
2. Create a new project group
3. Check any checkbox element
4. Notice that font of text related to the checkbox item is changed

**Expected behavior**
The font (font-weight) of the text shouldn't change on checking the checkbox.

**Additional context**
-

solid-maxim:
Fixes #554


illia-romanenko (approved):
LGTM - thanks @solid-maxim

PR #563, opened 2020-09-16
fix: Remove checkbox selection animation

### LINKED ISSUE #555: Remove checkbox selection animation
**Describe the bug**
We don't like the selection effect- (Default Material animation like InkWell) kind of glow around the box.
image

**To Reproduce**
1. Use any checkbox element
2. Notice that on selection default Splash effect applies

**Expected behavior**
No selection animation should be applied.

**Additional context**

solid-maxim:
Fixes #555

### DIFF
--- a/metrics/web/lib/project_groups/presentation/widgets/project_checkbox_list_tile.dart
+++ b/metrics/web/lib/project_groups/presentation/widgets/project_checkbox_list_tile.dart
@@ -26,7 +26,6 @@ class ProjectCheckboxListTile extends StatelessWidget {
     return Material(
       color: Colors.transparent,
       child: InkWell(
-        highlightColor: Colors.transparent,
         onTap: () => _toggleProjectCheckedStatus(context),
         child: Container(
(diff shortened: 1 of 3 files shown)


metrics/web/lib/project_groups/presentation/widgets/project_checkbox_list_tile.dart
```diff
@@ -26,6 +26,7 @@ class ProjectCheckboxListTile extends StatelessWidget {
     return Material(
       color: Colors.transparent,
       child: InkWell(
+        highlightColor: Colors.transparent,
```
vova-beloded, line 29: Maybe we can move this color to the MaterialTheme also?

vova-beloded (changes requested):
Great job, @solid-maxim! I've added one suggestion here, please, let me know what do you think.

vova-beloded (approved):
LGTM, thanks @solid-maxim!

PR #568, opened 2020-09-17
fix: Disable buttons highlights.

### LINKED ISSUE #547: Disable button highlights
**Describe the bug**
Disable highlight on pressing buttons like on the image below
image
Occurs in multiple places - should be fixed in all places.

**To Reproduce**
> Steps to reproduce the behavior:

1. Navigate to Project Groups page
2. Tap back (arrow) to return to the Dashboard
3. Notice under the arrow - circle highlight is displayed

**Expected behavior**
No circular highlight is displayed.

**Additional context**
-

solid-dimakoniaiev:
Fixes #547 

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/metrics/web/lib/auth/presentation/widgets/sign_in_option_button.dart
+++ b/metrics/web/lib/auth/presentation/widgets/sign_in_option_button.dart
@@ -21,6 +21,7 @@ class SignInOptionButton extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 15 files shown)


solid-vladlaba:
As we discussed in chat, let's check the solution with `HitTestBehavior` once more.

solid-vladlaba (approved):
I've checked all the clickable elements and seems like there is no ripple effect anymore - well done!

PR #569, opened 2020-09-17
fix: Add backdrop color for dialogs

### LINKED ISSUE #560: Modal dialog background overlay not matches Zeplin
**Describe the bug**
The color of the modal dialog background overlay should be copied from Zeplin on every modal popup:
image

**To Reproduce**

1. Navigate to Project Groups page
2. Press Add project group
3. Notice modal dialog shows
4. Observe that dialog overlay background color is different from Zeplin

**Expected behavior**
For every modal dialog set overlay background color from Zeplin

**Additional context**

solid-maxim:
Fixes #560 

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/metrics/web/lib/project_groups/presentation/widgets/add_project_group_card.dart
+++ b/metrics/web/lib/project_groups/presentation/widgets/add_project_group_card.dart
@@ -74,6 +74,8 @@ class AddProjectGroupCard extends StatelessWidget {
 
   /// Shows a [AddProjectGroupDialog] with an active project group.
   Future<void> _showProjectGroupDialog(BuildContext context) async {
+    final barrierColor =
+        MetricsTheme.of(context).projectGroupDialogTheme.barrierColor;
     final projectGroupsNotifier = Provider.of<ProjectGroupsNotifier>(
       context,
       listen: false,
@@ -84,6 +86,7 @@ class AddProjectGroupCard extends StatelessWidget {
     if (projectGroupsNotifier.projectGroupDialogViewModel == null) return;
 
     await showDialog(
+      barrierColor: barrierColor,
       context: context,
       child: AddProjectGroupDialog(),
     );
--- a/metrics/web/lib/project_groups/presentation/widgets/project_group_card.dart
+++ b/metrics/web/lib/project_groups/presentation/widgets/project_group_card.dart
@@ -59,7 +59,7 @@ class _ProjectGroupCardState extends State<ProjectGroupCard> {
                     mainAxisAlignment: MainAxisAlignment.spaceBetween,
                     children: <Widget>[
                       IconLabelButton(
-                        onPressed: () => _showEditProjectGroupDialog(context),
+                        onPressed: () => _showEditProjectGroupDialog(),
                         iconPadding: _buttonIconPadding,
                         icon: Image.network(
                           'icons/edit.svg',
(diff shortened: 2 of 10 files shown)


metrics/web/lib/project_groups/presentation/widgets/project_group_card.dart
```diff
@@ -144,6 +147,7 @@ class _ProjectGroupCardState extends State<ProjectGroupCard> {
     if (projectGroupsNotifier.deleteProjectGroupDialogViewModel == null) return;
 
     await showDialog(
+      barrierColor: barrierColor,
```
vova-beloded, line 150: Maybe we can extract the `showDialog` function call to the separate method like `_showProjectGroupDialog` to avoid code duplication?

metrics/web/lib/project_groups/presentation/widgets/add_project_group_card.dart
```diff
@@ -84,6 +86,7 @@ class AddProjectGroupCard extends StatelessWidget {
     if (projectGroupsNotifier.projectGroupDialogViewModel == null) return;
 
     await showDialog(
+      barrierColor: barrierColor,
```
vova-beloded, line 89: Should we test that the `barrierColor` is applied?

vova-beloded (changes requested):
Good job, @solid-maxim! I've added a couple of suggestions here, let me know what do you think.

metrics/web/lib/project_groups/presentation/widgets/project_group_card.dart
```diff
@@ -164,11 +164,27 @@ class _ProjectGroupCardState extends State<ProjectGroupCard> {
...
     );
 
     projectGroupsNotifier.resetProjectGroupDialogViewModel();
   }
+
+  /// Shows a project group dialog
+  /// and applies a barrier color from the metrics theme.
+  Future<void> _showProjectGroupDialog({
+    BuildContext context,
+    WidgetBuilder builder,
```
vova-beloded, line 179: Since this is a stateful widget, we can get the build context from the `context` getter of the `State` class, so we should avoid passing the context to the methods. Also, it seems like it will be better to pass the `Widget` instead of `WidgetBuilder`. What do you think?

metrics/web/test/project_groups/presentation/widgets/add_project_group_card_test.dart
```diff
@@ -137,6 +143,36 @@ void main() {
...
+        );
+
+        await tester.tap(find.byType(AddProjectGroupCard));
+        await tester.pumpAndSettle();
+
+        final modalBarrier = tester.widget<ModalBarrier>(
+          find.byType(ModalBarrier).last,
+        );
+
+        expect(modalBarrier.color, barrierColor);
```
vova-beloded, line 172: Would it be more reliable? 
```suggestion
        final barrierFinder = find.byWidgetPredicate(
          (widget) => widget is ModalBarrier && widget.color == barrierColor,
        );

        expect(barrierFinder, findsOneWidget);
```

vova-beloded (changes requested):
Well done, @solid-maxim! I've added a couple of suggestions, please, take a look.

vova-beloded (approved):
LGTM, thanks @solid-maxim!

PR #580, opened 2020-09-18
fix: Remove button shadows

### LINKED ISSUE #572: Remove button shadows
**Describe the bug**
Some buttons like Delete on Delete Project Group dialog are with shadow, 

image

**To Reproduce**

1. Open delete a project group dialog
2. Use the inspector to see that there is a shadow under the button

**Expected behavior**
Make sure that all relevant buttons don't contain shadow.

**Additional context**

solid-yuriiuvarov:
Fixes #572 

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/metrics/web/lib/common/presentation/button/theme/style/metrics_button_style.dart
+++ b/metrics/web/lib/common/presentation/button/theme/style/metrics_button_style.dart
@@ -18,7 +18,7 @@ class MetricsButtonStyle {
   const MetricsButtonStyle({
     this.color = Colors.blue,
     this.hoverColor = Colors.black12,
+    this.elevation = 0.0,
     this.labelStyle,
-    this.elevation,
   });
 }
--- a/metrics/web/lib/common/presentation/button/theme/style/metrics_button_style.dart
+++ b/metrics/web/lib/common/presentation/button/theme/style/metrics_button_style.dart
@@ -15,6 +15,10 @@ class MetricsButtonStyle {
   final double elevation;
 
   /// Creates a new instance of the [MetricsButtonStyle].
+  ///
+  /// The [color] default value is [Colors.blue].
+  /// The [hoverColor] default value is [Colors.black12].
+  /// The [elevation] default value is `0.0`.
   const MetricsButtonStyle({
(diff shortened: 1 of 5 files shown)


metrics/web/lib/common/presentation/metrics_theme/model/dark_metrics_theme_data.dart
```diff
@@ -201,6 +201,7 @@ class DarkMetricsThemeData extends MetricsThemeData {
             buttonAttentionLevel: MetricsButtonAttentionLevel(
               positive: MetricsButtonStyle(
                 color: ColorConfig.primaryColor,
+                elevation: 0.0,
```
illia-romanenko, line 204: Should we just make elevation default value `0` in MetricsButtonStyle? As by default, I think elevation is 0 and most likely will be 0 for all of the cases. In that case, we can remove all the places where you explicitly say it's 0.0. It might be also worth considering removing elevation from MetricsButtonStyle constructor - is most likely we'll not have elevation here at all.

illia-romanenko (changes requested):
Great job - I've added one comment - please let me know what you think!

solid-vladlaba (approved):

metrics/web/lib/common/presentation/button/theme/style/metrics_button_style.dart
```diff
@@ -18,7 +18,7 @@ class MetricsButtonStyle {
   const MetricsButtonStyle({
     this.color = Colors.blue,
     this.hoverColor = Colors.black12,
+    this.elevation = 0.0,
```
illia-romanenko, line 21: I think our documentation approach is to specify all default values? In this case we seem to lack them.

illia-romanenko (changes requested):
Looks great - thanks! I think we should update documentation as well and then it should be good to go.

illia-romanenko (approved):
LGTM - thanks @solid-dimakoniaiev @solid-yuriiuvarov !

illia-romanenko:
@mergifyio refresh

PR #582, opened 2020-09-18
fix: On user logout clear saved project groups on the Dashboard

### LINKED ISSUE #578: On user logout clear saved project groups on the Dashboard
**Describe the bug**
User can see cached project groups data:
image

**To Reproduce**

1. Sign in with user/password (Password sign-in)
2. Notice that project group dropdown on Dashboard is populated
3. Logout
4. Stay on the same page without refresh and sign in with Google auth user that has associated email domain that is not allowed (allowed_email_domains)
5. Notice project groups are displayed still because they were cached in the app (if you refresh the page on step 4 - it will not display any groups)

**Expected behavior**
On user logout - the project groups cache is cleared

**Additional context**

solid-yuriiuvarov:
Fixes #578 

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/metrics/web/test/dashboard/presentation/state/project_metrics_notifier_test.dart
+++ b/metrics/web/test/dashboard/presentation/state/project_metrics_notifier_test.dart
@@ -353,6 +353,39 @@ void main() {
     );
 
     test(
+      ".resetSelectedProjectGroup() resets the selected project group",
+      () async {
+        const projectGroupId = 'id';
+        final projectGroups = [
+          ProjectGroupModel(
+            id: projectGroupId,
+            name: 'name',
+            projectIds: UnmodifiableListView([]),
+          ),
+          ProjectGroupModel(
+            id: 'id2',
+            name: 'name1',
+            projectIds: UnmodifiableListView([]),
+          ),
+        ];
+
+        projectMetricsNotifier.setProjectGroups(projectGroups);
+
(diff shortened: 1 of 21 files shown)


illia-romanenko (changes requested):
@solid-yuriiuvarov @solid-dimakoniaiev  you are only cleaning selectedProjectGroup, but you should also clean the list of project groups.

metrics/web/test/dashboard/presentation/state/project_metrics_notifier_test.dart
```diff
@@ -669,6 +669,34 @@ void main() {
         );
       },
     );
+
+    test(
+      ".selectProjectGroup() resets project group dropdown items and selected project group to null if project groups are null",
```
solid-vladlaba, line 674:
```suggestion
      ".setProjectGroups() resets project group dropdown items and selected project group to null if project groups are null",
```
And seems like this test case lacks formatting.

solid-vladlaba (changes requested):
Well done, @solid-yuriiuvarov @solid-dimakoniaiev! I've noticed one small mistake in a method name in tests. But overall everything is good!

illia-romanenko (approved):
LGTM - thanks @solid-yuriiuvarov @solid-dimakoniaiev !

PR #585, opened 2020-09-21
fix: Remove an additional scroll stretching effect

### LINKED ISSUE #566: Remove an additional scroll stretching effect
**Describe the bug**
Default material scrolling effect applies but we don't want it to be there.
 
image

**To Reproduce**

1. Navigate to the dashboard page
2. Scroll project metrics down
3. Notice scrolling effect

**Expected behavior**
No scrolling effect should be applied across the project (for every scroll there shouldn't be any effect). Should we create some common widget for that?

**Additional context**

solid-dimakoniaiev:
Fixes #566

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/metrics/web/lib/main.dart
+++ b/metrics/web/lib/main.dart
@@ -125,4 +125,13 @@ class _ScrollBehaviour extends ScrollBehavior {
   ScrollPhysics getScrollPhysics(BuildContext context) {
     return const ClampingScrollPhysics();
   }
+
+  @override
+  Widget buildViewportChrome(
+    BuildContext context,
+    Widget child,
+    AxisDirection axisDirection,
+  ) {
+    return child;
+  }
 }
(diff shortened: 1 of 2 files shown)


metrics/web/lib/main.dart
```diff
@@ -112,3 +118,15 @@ class _MetricsAppState extends State<MetricsApp> {
     );
   }
 }
+
+/// A [ScrollBehavior] that disable the stretching effect.
+class _ScrollBehaviour extends ScrollBehavior {
```
illia-romanenko, line 123:
```suggestion
class MetricsScrollBehavior extends ScrollBehavior {
```
I think it also would be beneficial to move it to the `common/presentation/widgets` to make the `main.dart` more simple and to match Flutter's `ScrollBehavior` location (`src/widgets/scroll_configuration.dart`)

illia-romanenko (changes requested):
Thanks, @solid-dimakoniaiev @solid-yuriiuvarov - looks great now!

I've added some structure suggestion - please review and let me know what you think.

The site looks more and more polished each day - thanks!

illia-romanenko (approved):
LGTM - thanks @solid-yuriiuvarov @solid-dimakoniaiev !

illia-romanenko:
@mergifyio refresh

PR #586, opened 2020-09-21
fix: Email domain validation error message and google sign-in pop-up

### LINKED ISSUE #470: Add Google Sign-in domain allow list

As an administrator, I want to make sure that only users from certain organizations can log in to the app so I can make sure that data protected.

It's related to data protection and access control.

As a user, I can only login via Google Sign-in if my email domain is in the allow list. Allow list should be configurable via Firebase and we need to make sure that the data is protected reliably as Firebase appears to not provide domain validation out of the box, so it most likely will be a custom implementation using security rules.

**Describe alternatives you've considered**
Use cloud functions to generate tokens.

**Additional context**
All relevant security tests should be added for Firebase security rules.

solid-maxim:
Fixes #470

### DIFF
--- a/metrics/web/lib/auth/presentation/models/auth_error_message.dart
+++ b/metrics/web/lib/auth/presentation/models/auth_error_message.dart
@@ -27,6 +27,8 @@ class AuthErrorMessage {
         return AuthStrings.unknownErrorMessage;
       case AuthErrorCode.googleSignInError:
         return AuthStrings.googleSignInErrorOccurred;
+      case AuthErrorCode.notAllowedEmailDomain:
+        return AuthStrings.notAllowedEmailDomainErrorMessage;
       default:
         return null;
     }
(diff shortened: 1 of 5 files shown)


metrics/web/lib/auth/presentation/models/auth_error_message.dart
```diff
@@ -27,6 +27,8 @@ class AuthErrorMessage {
         return AuthStrings.unknownErrorMessage;
       case AuthErrorCode.googleSignInError:
         return AuthStrings.googleSignInErrorOccurred;
+      case AuthErrorCode.notAllowedEmailDomain:
```
vova-beloded, line 30: Could you please add more tests to the auth error message tests to cover all available mapping cases?

vova-beloded (changes requested):
Great job, @solid-maxim! I've added one suggestion here - please take a look.

vova-beloded (approved):
LGTM, thanks @solid-maxim!

PR #614, opened 2020-09-23
feat: Login Page theme matches OS theme

### LINKED ISSUE #593: Login page theme should match OS theme

As a user, I want to have matching with my OS site theme so I can have a smooth experience.

When user navigates to the login page theme should match OS theme, if user changes OS theme - then app theme should be switched (on the login page not in general).

We can introduce observer: https://stackoverflow.com/questions/57475630/detect-if-ios13-dark-mode-is-enabled-from-flutter-dart/58227507#58227507 - just need to make sure that it works on the web or use Favicon class approach.

Move Favicon.setup -> main().

solid-daniyilyevtyushkin:
Fixes #593 

## Checklist

- [x] Documented
- [x] Unit tested
- [x] Component tested
- [x] Driver tested

### DIFF
--- a/metrics/web/lib/auth/presentation/pages/login_page.dart
+++ b/metrics/web/lib/auth/presentation/pages/login_page.dart
@@ -12,7 +12,7 @@ import 'package:provider/provider.dart';
 
 /// Shows the authentication form to sign in.
 class LoginPage extends StatefulWidget {
-  /// Creates a new instance of [LoginPage].
+  /// Creates a new instance of the [LoginPage].
   const LoginPage({
     Key key,
   }) : super(key: key);
--- a/metrics/web/lib/auth/presentation/pages/login_page.dart
+++ b/metrics/web/lib/auth/presentation/pages/login_page.dart
@@ -58,9 +58,7 @@ class _LoginPageState extends State<LoginPage> with WidgetsBindingObserver {
     final platformBrightness =
         WidgetsBinding.instance.window.platformBrightness;
 
-    final isDark = platformBrightness == Brightness.dark;
-
-    _themeNotifier.setTheme(isDark: isDark);
+    _themeNotifier.setTheme(platformBrightness);
   }
 
   /// Navigates to the dashboard screen once the user becomes logged in.
--- a/metrics/web/lib/auth/presentation/pages/login_page.dart
+++ b/metrics/web/lib/auth/presentation/pages/login_page.dart
@@ -41,7 +41,7 @@ class _LoginPageState extends State<LoginPage> with WidgetsBindingObserver {
 
     WidgetsBinding.instance.addObserver(this);
     WidgetsBinding.instance.addPostFrameCallback((_) {
-      _changeThemeAccordingOperatingSystemTheme();
+      _handlePlatformBrightness();
     });
 
     super.initState();
@@ -49,12 +49,12 @@ class _LoginPageState extends State<LoginPage> with WidgetsBindingObserver {
 
   @override
   void didChangePlatformBrightness() {
-    _changeThemeAccordingOperatingSystemTheme();
+    _handlePlatformBrightness();
     super.didChangePlatformBrightness();
   }
 
   /// Changes the current theme according to the operating system's theme.
-  void _changeThemeAccordingOperatingSystemTheme() {
+  void _handlePlatformBrightness() {
     final platformBrightness =
         WidgetsBinding.instance.window.platformBrightness;
 
--- a/metrics/web/lib/auth/presentation/pages/login_page.dart
+++ b/metrics/web/lib/auth/presentation/pages/login_page.dart
@@ -12,6 +12,11 @@ import 'package:provider/provider.dart';
 
 /// Shows the authentication form to sign in.
 class LoginPage extends StatefulWidget {
+  /// Creates a new instance of [LoginPage].
+  const LoginPage({
+    Key key,
+  }) : super(key: key);
+
   @override
   _LoginPageState createState() => _LoginPageState();
 }
--- a/metrics/web/lib/auth/presentation/pages/login_page.dart
+++ b/metrics/web/lib/auth/presentation/pages/login_page.dart
@@ -1,6 +1,7 @@
 import 'package:flutter/material.dart';
 import 'package:metrics/auth/presentation/state/auth_notifier.dart';
 import 'package:metrics/auth/presentation/widgets/auth_form.dart';
+import 'package:metrics/common/presentation/metrics_theme/state/theme_notifier.dart';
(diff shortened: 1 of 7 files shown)


metrics/web/lib/common/presentation/metrics_theme/state/theme_notifier.dart
```diff
@@ -14,4 +21,10 @@ class ThemeNotifier extends ChangeNotifier {
     _isDark = !_isDark;
     notifyListeners();
   }
+
+  /// Sets the [_isDark] value according to the given [isDark] value.
+  void setTheme({bool isDark}) {
+    _isDark = isDark;
```
solid-vladlaba, line 27: As I remember we should change the theme mode to the OS selected only on the Login page. But I'm wondering should we handle the following scenario: 
1. A user logs in to the Metrics with the light theme on the Login page (*OS: light*, *app: light*).
2. A user changes the theme mode to the dark on the user menu (*OS: light*, *app: dark*).
3. A user logs out of the application and is moved to the Login page  (*OS: light*, *app: ???*).

What the mode should be selected instead of `???`. According to the implementation in this PR, the mode will be light as the OS states. 

cc @illia-romanenko

metrics/web/lib/common/presentation/metrics_theme/state/theme_notifier.dart
```diff
@@ -14,4 +21,10 @@ class ThemeNotifier extends ChangeNotifier {
     _isDark = !_isDark;
     notifyListeners();
   }
+
+  /// Sets the [_isDark] value according to the given [isDark] value.
+  void setTheme({bool isDark}) {
```
solid-vladlaba, line 26: I suppose we should handle the case when the given `isDark` value is `null`.

metrics/web/lib/auth/presentation/pages/login_page.dart
```diff
@@ -11,25 +12,57 @@ import 'package:provider/provider.dart';
...
   }
 
+  @override
+  void didChangePlatformBrightness() {
+    _changeThemeAccordingOperatingSystemTheme();
+    super.didChangePlatformBrightness();
+  }
+
+  /// Changes the current theme according to the operating system's theme.
+  void _changeThemeAccordingOperatingSystemTheme() {
```
solid-vladlaba, line 57:
```suggestion
  void _handlePlatformBrightness() {
```

metrics/web/lib/common/presentation/metrics_theme/state/theme_notifier.dart
```diff
@@ -4,7 +4,14 @@ import 'package:flutter/foundation.dart';
 ///
 /// Stores and provides currently selected theme type.
 class ThemeNotifier extends ChangeNotifier {
-  bool _isDark = true;
+  bool _isDark;
+
+  /// Creates a new [ThemeNotifier] instance.
+  ///
+  /// [_isDark] defaults to `true`.
```
solid-vladlaba, line 11:
```suggestion
  /// If the given [isDark] is `null`, the `true` is used.
```

metrics/web/test/auth/presentation/pages/login_page_test.dart
```diff
@@ -129,6 +152,9 @@ void main() {
 
 /// A testbed widget, used to test the [LoginPage] widget.
 class _LoginPageTestbed extends StatelessWidget {
+  /// A [GlobalKey] needed to get the current context of the [LoginPage].
+  static final GlobalKey childKey = GlobalKey();
```
solid-vladlaba, line 156:
```suggestion
  /// A [GlobalKey] for the [LoginPage] to use in tests.
  static final GlobalKey loginKey = GlobalKey();
```

And I think it would be better to pass the key as a parameter in the constructor.

metrics/web/lib/common/presentation/metrics_theme/state/theme_notifier.dart
```diff
@@ -4,7 +4,14 @@ import 'package:flutter/foundation.dart';
...
+  ThemeNotifier({
+    bool isDark,
```
solid-vladlaba, line 13: I'm not sure we can initialize the notifier with such a parameter.

metrics/web/test/auth/presentation/pages/login_page_test.dart
```diff
@@ -63,6 +64,28 @@ void main() {
       },
     );
 
+    testWidgets(
+      "applies the theme brightness that corresponds the operating system's brightness",
```
solid-vladlaba, line 68: Can we also test that the Login page listens on the platform brightness somehow? I mean, the test case checking that if the platform brightness changes, the theme mode in the notifier changes as well.

metrics/web/test/common/presentation/metrics_theme/state/theme_notifier_test.dart
```diff
@@ -1,10 +1,36 @@
...
 
+// https://github.com/software-platform/monorepo/issues/140
+// ignore_for_file: prefer_const_constructors, avoid_redundant_argument_values
+
 void main() {
   group("ThemeNotifier", () {
-    test(".changeTheme() changes the theme", () {
+    test(
+        "creates an instance with the default is dark value if the is dark parameter is not specified",
+        () {
```
solid-vladlaba, line 11: It seems like this file lacks formatting.

metrics/web/test/common/presentation/metrics_theme/state/theme_notifier_test.dart
```diff
@@ -13,11 +39,28 @@ void main() {
...
-      final themeNotifier = ThemeNotifier();
+      final themeNotifier = ThemeNotifier(isDark: true);
 
       themeNotifier.addListener(expectAsync0(() {}));
 
       themeNotifier.changeTheme();
     });
+
+    test(
+        ".setTheme() changes the theme according to the given is dark value and notifies listeners",
```
solid-vladlaba, line 50: Seems like we should divide this test case into two separate as for the `changeTheme` method.

solid-vladlaba (changes requested):
Well done, @solid-daniyilyevtyushkin! I've added a couple of suggestions below - please let me know what you think!

A tremendous solution with the `WidgetsBindingObserver`! 👍

metrics/web/lib/common/presentation/metrics_theme/state/theme_notifier.dart
bootstraponline: I think light is correct, we decided for the login page to follow the OS theme.

solid-vladlaba: Cool, thanks!

solid-vladlaba (approved):
LGTM, thanks @solid-daniyilyevtyushkin!:rocket:

PR #631, opened 2020-09-28
fix: Long dropdown list have height of 5.5 items

### LINKED ISSUE #602: Long dropdown list should have height of 5.5 items

As a user, I want to know that there are more elements than I see so I can understand that dropdown list is scrollable.
image

`The list is a bit too short - The idea is to show 5 and a half item - The user will notice that there are more items that 5 and the list is scrollable. Now list shows 5 items without half so It looks there are 5 items at all.`

If there are more than 5 elements - the height should be 5.5 elements, if 5 or less - than it should take the actual space.

**Describe alternatives you've considered**

**Additional context**

solid-tetianamovlian:
Fixes #602

### DIFF
--- a/metrics/web/test/base/presentation/widgets/dropdown_menu_test.dart
+++ b/metrics/web/test/base/presentation/widgets/dropdown_menu_test.dart
@@ -143,7 +143,7 @@ void main() {
     );
 
     testWidgets(
-      "uses a zero padding as a default if the list padding parameter is null",
+      "applies a zero list padding if the given list padding is null",
       (tester) async {
         await tester.pumpWidget(const _DropdownMenuTestbed(listPadding: null));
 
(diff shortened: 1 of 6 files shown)


metrics/web/test/base/presentation/widgets/dropdown_menu_test.dart
```diff
@@ -141,6 +142,23 @@ void main() {
       },
     );
 
+    testWidgets(
+      "uses a zero padding as a default if the list padding parameter is null",
```
solid-vovabeloded, line 146: Would it be more clear? 
```suggestion
      "applies a zero list padding if the give list padding is null",
```

solid-vovabeloded (changes requested):
Well done, @solid-maxim & @solid-tetianamovlian! I've added a couple of suggestions - please take a look.

solid-vovabeloded (approved):

PR #640, opened 2020-10-01
feat: User profile data

### LINKED ISSUE #621: User Profile - Data layer

As a developer, I want to make sure that User Profile data can be accessed from Firebase/Firestore so I can implement User Profile related functionality.

No ability to store User Profile.

1. Data can be serialized and deserialized from Firestore/JSON
2. User can create/read/update User Profile
3. Implemented according to Architecture document.

solid-maxim:
Fixes #621 

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/metrics/web/lib/auth/data/model/user_profile_data.dart
+++ b/metrics/web/lib/auth/data/model/user_profile_data.dart
@@ -1,3 +1,4 @@
+import 'package:meta/meta.dart';
 import 'package:metrics/auth/domain/entities/theme_type.dart';
 import 'package:metrics/auth/domain/entities/user_profile.dart';
 import 'package:metrics_core/metrics_core.dart';
@@ -6,8 +7,8 @@ import 'package:metrics_core/metrics_core.dart';
 class UserProfileData extends UserProfile implements DataModel {
   /// Creates the [UserProfileData] with the given [id] and [selectedTheme].
   UserProfileData({
-    String id,
-    ThemeType selectedTheme,
+    @required String id,
+    @required ThemeType selectedTheme,
   }) : super(id: id, selectedTheme: selectedTheme);
 
   /// Creates the [UserProfileData] using the [json] and the [documentId].
--- a/metrics/web/lib/auth/data/model/user_profile_data.dart
+++ b/metrics/web/lib/auth/data/model/user_profile_data.dart
@@ -1,0 +1,34 @@
+import 'package:metrics/auth/domain/entities/theme_type.dart';
+import 'package:metrics/auth/domain/entities/user_profile.dart';
+import 'package:metrics_core/metrics_core.dart';
+
+/// A [DataModel] that represents the [UserProfile] entity.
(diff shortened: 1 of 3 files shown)


metrics/web/lib/auth/data/repositories/firebase_user_repository.dart
```diff
@@ -107,11 +112,51 @@ class FirebaseUserRepository implements UserRepository {
...
+        .snapshots()
+        .map(
+          (snapshot) =>
+              UserProfileData.fromJson(snapshot.data, snapshot.documentID),
+        )
+        .handleError((_) {
+      throw const PersistentStoreException(
+        code: PersistentStoreErrorCode.unknown,
+      );
+    });
```
solid-vovabeloded, line 128:
```suggestion
    return _firestore.collection('user_profiles').document(id).snapshots().map(
      (snapshot) {
        return UserProfileData.fromJson(
          snapshot.data,
          snapshot.documentID,
        );
      },
    ).handleError((_) {
      throw const PersistentStoreException(
        code: PersistentStoreErrorCode.unknown,
      );
    });
```
?

metrics/web/lib/auth/data/model/user_profile_data.dart
```diff
@@ -0,0 +1,35 @@
...
+  /// Returns `null` if the given [json] is `null`.
+  factory UserProfileData.fromJson(
+    Map<String, dynamic> json,
+    String documentId,
+  ) {
+    if (json == null) return null;
+
+    return UserProfileData(
+      id: documentId,
+      selectedTheme: json['selectedTheme'] as ThemeType,
```
solid-vovabeloded, line 25: Since we can't store the `enum` values in the database directly, we are storing them as a string representation, so we should map it here from `String` to `enum` value. You can find an example of mapping the enum in the [build_data_deserializer.dart](https://github.com/platform-platform/monorepo/blob/master/metrics/web/lib/dashboard/data/deserializer/build_data_deserializer.dart) file.

metrics/web/lib/auth/data/model/user_profile_data.dart
```diff
@@ -0,0 +1,35 @@
...
+    );
+  }
+
+  @override
+  Map<String, dynamic> toJson() {
+    return {
+      'selectedTheme': selectedTheme,
```
solid-vovabeloded, line 32: We should store the `enum` as a `String` in the database, so it should look like this: 
```suggestion
      'selectedTheme': selectedTheme.toString(),
```
or 
```suggestion
      'selectedTheme': '$selectedTheme',
```

solid-vovabeloded (changes requested):
Well done, @solid-maxim & @solid-tetianamovlian! I've added a couple of suggestions here - please take a look.

solid-vovabeloded (approved):
LGTM, thanks @solid-maxim & @solid-tetianamovlian!

PR #663, opened 2020-10-05
feat: Performance chart group data by day.

### LINKED ISSUE #645: Performance chart group data by day

As a user, I want to see compact and meaningful performance data in case there were many builds during the last 7 days so I can understand it better.

If there were many builds during the last week chart displays too much data so it becomes unreadable.

- Display mean (average) build time for the day and display data for the last 7 days. 
image
so usually there will be 7 points with the average duration for that day
- If there were no builds for the last 7 days - show "No data" performance chart. 
image
- If there were no builds in one day but there were builds during other days in the last 7 days - display it as 0 for the day that there were no builds (for example that usually should happen on the weekends)

solid-dimakoniaiev:
Fixes #645 

## Checklist

- [x] Documented
- [x] Unit tested


solid-vladlaba (approved):
I've improved an algorithm and test a bit to make it more readable and clear. Please review the improvements and let me know what you think!

PR #665, opened 2020-10-06
fix: Circle percentage widget with null value

### LINKED ISSUE #646: Dashboard filters show data for empty group
**Describe the bug**
If the user filters the data, wrong data is presented for the empty group:
image

**To Reproduce**
1. Open test site
2. Filter by "32. Project next to 31." project group
3. Notice that for the empty group for Stability and Coverage is displaying some progress but it should be empty like it's when the user doesn't apply filter.
image

**Expected behavior**
Empty group data should always be empty.
image

Suggestion: try to use didUpdateWidget to solve this.

solid-vovabeloded:
Fixes #646 

## Checklist

- [x] Documented
- [x] Component tested


solid-vladlaba (approved):

PR #669, opened 2020-10-06
fix: Fix diagram links in user profile theme design document

solid-vovabeloded:


illia-romanenko (approved):

PR #685, opened 2020-10-08
fix: Move metrics text style to an appropriate package

solid-vovabeloded:


solid-vladlaba (approved):

PR #705, opened 2020-10-14
fix: Update chromedriver version to match latest Google Chrome

solid-vovabeloded:


solid-vladlaba (approved):

PR #706, opened 2020-10-14
fix: Fix performance and builds metrics loading

solid-vovabeloded:


solid-vladlaba (approved):

PR #711, opened 2020-10-15
fix: Fix circle percentage value text displaying

solid-vovabeloded:

### DIFF
--- a/metrics/web/test_driver/common/config/driver_tests_config.dart
+++ b/metrics/web/test_driver/common/config/driver_tests_config.dart
@@ -15,5 +15,5 @@ class DriverTestsConfig {
   /// Will be used to download the chrome driver if there are no 'chromedriver'
   /// file will be found in working directory
   static const String chromeDriverDownloadUrl =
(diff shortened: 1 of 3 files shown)


metrics/web/test_driver/common/config/driver_tests_config.dart
```diff
@@ -15,5 +15,5 @@ class DriverTestsConfig {
   /// Will be used to download the chrome driver if there are no 'chromedriver'
   /// file will be found in working directory
   static const String chromeDriverDownloadUrl =
-      'https://chromedriver.storage.googleapis.com/86.0.4240.22/chromedriver_mac64.zip';
+      'https://chromedriver.storage.googleapis.com/85.0.4183.38/chromedriver_mac64.zip';
```
bootstraponline, line 18: Is this intentional?

solid-vovabeloded: Yep, it seems like GitHub Actions downgraded the Google Chrome version, so we need a lower version of the chrome driver to be able to run the driver tests.

solid-vladlaba (approved):

PR #777, opened 2020-11-04
fix: URLs in models must link to Github web

solid-vladlaba:


solid-vovabeloded (approved):
LGTM, thanks @solid-vladlaba!

PR #795, opened 2020-11-10
fix: Rename firebase API key to the firebase public API key.

### LINKED ISSUE #790: Rename firebase_api_key -> firebase_public_api_key
**Description**
As we want to make config files more readable it would make sense to rename firebase_api_key -> firebase_public_api_key to make sure that users understand that this is a public key and not a private key - thus can be entered freely and stored in the version control tool.

**Acceptance Criteria**
- [x] CI Integrations: change destination config, parser, etc.. firebase_api_key -> firebase_public_api_key
- [x] Change existing configurations in ".metrics"

solid-dimakoniaiev:
Fixes #790 

## Checklist

- [x] Documented
- [x] Unit tested


solid-vladlaba (approved):

PR #799, opened 2020-11-10
fix: Override the `coverage` dependency to use the latest version from the GitHub

solid-vovabeloded:


illia-romanenko (approved):

illia-romanenko:
@mergifyio refresh

PR #800, opened 2020-11-10
fix: Fix test_coverage dependency to be compatible with latest coverage package

solid-vovabeloded:


illia-romanenko (approved):

illia-romanenko:
@mergifyio refresh

PR #801, opened 2020-11-10
fix: Update Ci Integrations link to point to latest version

solid-vovabeloded:


illia-romanenko (approved):
LGTM - thanks @solid-vovabeloded,  look forward to setup normal builds tomorrow!

PR #844, opened 2020-11-26
feat: Improve test driver to be able to run in docker

solid-vovabeloded:


illia-romanenko (approved):
LGTM - thanks @solid-vovabeloded. Love it when we reduce the codebase while using the new functionality.

PR #858, opened 2020-12-01
fix: Rename info icon

solid-dimakoniaiev:


solid-vovabeloded (approved):
LGTM, thanks @solid-dimakoniaiev & @solid-tetianamovlian!

PR #860, opened 2020-12-01
fix: Fix SyncError test group name

solid-vovabeloded:


solid-vladlaba (approved):

PR #894, opened 2020-12-14
fix: Change sync error message

solid-vladlaba:


illia-romanenko (approved):
LGTM - thanks @solid-vladlaba !

illia-romanenko:
@mergifyio refresh

PR #922, opened 2020-12-18
fix: Fix sending slack notifications in Metrics Integration Actions

solid-vovabeloded:


illia-romanenko (approved):

PR #942, opened 2020-12-28
fix: Fix resetting project group dialog view models

solid-vovabeloded:
Fixes #

## Checklist

- [x] Documented
- [x] Unit tested
- [x] Component tested


illia-romanenko (approved):
LGTM - thanks @solid-vovabeloded. Thanks for explaining several points and fixes according to Effective Dart!

PR #976, opened 2021-01-15
fix: Environment Variable substitution in actions

solid-vladlaba:
Fixes #


illia-romanenko (approved):
LGTM - thanks @solid-vladlaba.

PR #991, opened 2021-01-20
build: Simplify Firebase hosting targets

### LINKED ISSUE #988: Simplify firebase target configuration
**Description**
Currently, we have to separate targets in firebase config:

But since they are going to be the same - we want to simplify configuration and have only one configuration - "metrics" in `.firebaserc` and `firebase.json`
**Acceptance Criteria**
- [ ] one target is used in firebase configurations
- [ ] Github actions updated to support those changes.

solid-daniyilyevtyushkin:
Fixes #988

### DIFF
--- a/.github/workflows/web_actions.yml
+++ b/.github/workflows/web_actions.yml
@@ -108,7 +108,9 @@ jobs:
         working-directory: metrics/web/
 
       - name: Deploy HTML version to Firebase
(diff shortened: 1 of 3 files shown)


solid-vovabeloded (approved):
LGTM, thanks @solid-daniyilyevtyushkin & @solid-yevheniiashchypeleva!

PR #994, opened 2021-01-20
fix: Large builds per week numbers

### LINKED ISSUE #992: Builds per week should support larger numbers
**Describe the bug**
Builds per week should support larger numbers, larger than 1000.

**To Reproduce**

1. Set build per week more than 1000
2. Observe that only 100 is displayed

**Expected behavior**
The full number should be displayed but with reduced font size in such a case. Make sure that 2 digit builds are pixel-perfect matching the designs.

solid-daniyilyevtyushkin:
Fixes #992

## Checklist

- [x] Documented
- [x] Component tested
- [x] Ensure the 2-digit numbers are pixel-perfect

Снимок экрана 2021-01-20 в 16 30 53


solid-vovabeloded (approved):
LGTM, thanks @solid-daniyilyevtyushkin & @solid-yevheniiashchypeleva! Thanks for improving the existing codebase!

PR #1010, opened 2021-01-26
feat: CI integrations Logger improvements

### LINKED ISSUE #995: Metrics CLI: investigate logger design improvement to solve "prefix methods"
**Description**
We want to improve Metrics CLI Logger design to init instance with the prefix (using runtimeType?) so that we don't have to implement such methods for every class:

image

**Acceptance Criteria**
- [ ] Design updated with an approach that solves "prefix methods" or arguments presented against it

cc: @solid-vovabeloded

solid-vladlaba:
Fixes #995

## Checklist

- [x] Documented
- [x] Unit tested
- [x] Component tested

### DIFF
--- a/metrics/ci_integrations/test/cli/logger/logger_test.dart
+++ b/metrics/ci_integrations/test/cli/logger/logger_test.dart
@@ -194,4 +194,6 @@ void main() {
   });
 }
 
+/// A class stub to use in tests as [Logger.sourceClass] and ensure
+/// this source is unique within tests.
 class _ClassStub {}
--- a/metrics/ci_integrations/test/cli/logger/logger_test.dart
+++ b/metrics/ci_integrations/test/cli/logger/logger_test.dart
@@ -1,65 +1,197 @@
 import 'package:ci_integration/cli/logger/logger.dart';
-import 'package:intl/intl.dart';
+import 'package:ci_integration/cli/logger/manager/logger_manager.dart';
 import 'package:mockito/mockito.dart';
 import 'package:test/test.dart';
 
-import '../test_util/mock/io_sink_mock.dart';
+import '../test_util/mock/logger_factory_mock.dart';
+import '../test_util/mock/logger_writer_mock.dart';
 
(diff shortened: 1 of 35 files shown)


metrics/ci_integrations/test/cli/logger/logger_test.dart
```diff
@@ -1,65 +1,197 @@
...
+          writer: writerMock,
+          verbose: testVerbose,
+        ),
+        returnsNormally,
+      );
     });
 
     test(
-      ".logMessage() logs the given message to the message sink",
+      ".forClass() returns a new logger if LoggerManager does not contain any for the given class type",
```
solid-vovabeloded, line 94: Maybe we can just test that this method delegates to the `LoggerManager.getLogger()`? It seems like we've already tested that the `getLogger()` works correctly in `LoggerManager` tests.

metrics/ci_integrations/test/cli/logger/logger_test.dart
```diff
@@ -1,65 +1,197 @@
...
+          (log) => matchesPrefixes(log, '$testSourceClass', message),
+        );
+
+        verify(writerMock.write(argThat(logMatcher))).called(1);
       },
     );
   });
 }
+
+class _ClassStub {}
```
solid-vovabeloded, line 197: It seems like this class lacks documentation. Please, go through all tests and apply this suggestion.

solid-vovabeloded (changes requested):
Great job, @solid-vladlaba! I've added a couple of suggestions here - please take a look. Overall, the new logger looks great - thanks for improving it!

metrics/ci_integrations/test/cli/logger/logger_test.dart
solid-vladlaba: Resolve this as we discussed via chat

solid-vovabeloded (approved):
LGTM, thanks @solid-vladlaba!

PR #1036, opened 2021-02-01
docs: Store database metadata

### LINKED ISSUE #1001: Design store app/db version
**Description**
For future upgrade purposes store known app/db version so the app could react to the changes in it and ask user to refresh the page if app build version is different from the app/db version.

**Acceptance Criteria**
- [ ] Prepare document with investigation results of app/db version approach

**Additional context**

We should probably block old apps from accessing newer dbs.

solid-vovabeloded:
Fixes #1001 

## Checklist

- [x] Documented


metrics/docs/01_storing_database_metadata.md
```diff
@@ -0,0 +1,115 @@
...
+>   - metadata
+>       - databaseVersion: String
+>       - isUpdating : bool
+
+
+The `metadata` collection should have the following security rules: 
+
+- No one can write (update, delete, create) documents in this collection.
+- Anybody can read documents from this collection. 
+
```
illia-romanenko, line 57:
```suggestion

Firebase Admin SDK should be used to update the data.
```

metrics/docs/01_storing_database_metadata.md
```diff
@@ -0,0 +1,115 @@
...
+Also, the `version` file will simplify the process of getting the latest version during database deployment/updating to set the current database version.
+
+## Making things work
+> Explain the process of updating the Metrics applications.
+
+Once we have a database version in the Firestore database and an application version, we should block the Metrics applications from updating the database records when the database is updating or its version is not compatible with the current application version.
+
+Let's consider the process of blocking the Metrics Web Application: 
+
+### Metrics Web Application
```
illia-romanenko, line 93: Should we add some notes about CLI like `ci_integrations` how it should use version, etc...?

illia-romanenko (changes requested):
Great job @solid-vovabeloded! This is a complex problem we need to solve and I see an excellent solution to it. I've added a couple of suggestions - please let me know what you think. Loved how easy it's to read it and get the idea - well done!

PR #1044, opened 2021-02-01
feat: jenkins pipelines added

jdzsz:


.jenkins/pipelines/coverage_converter.groovy
```diff
@@ -0,0 +1,76 @@
...
+                dir('metrics/ci_integrations'){
+                    sh 'pub get'
+                }
+                script{
+                    isTested = false
+                }
+            }
+        }
+        stage('Parallel Stage') {
+                    parallel {
```
solid-vovabeloded, line 31: Should we add a [failFast true](https://www.jenkins.io/doc/book/pipeline/syntax/#parallel) flag here to fail the parallel build once the `dartanalyzer ` or `test_coverage` fails? Please, apply this suggestion to all pipelines that contains `parallel` step.
 ```suggestion
        stage('Parallel Stage') {
                    failFast true
                    parallel {
```

File: .jenkins/pipelines/web.groovy [hunk 2]
```diff
@@ -0,0 +1,80 @@
...
+
+                    }
+                }
+        stage('Convert coverage report') {
+            steps {
+                waitUntil{
+                    script{
+                        return isTested
+                    }
+                }
```
solid-vovabeloded, line 61: Can we remove `waitUntil` here since this stage goes after `Parallel Stage` and will be run only when it finishes?

.jenkins/pipelines/web.groovy
```diff
@@ -0,0 +1,80 @@
...
+                dir('metrics/web'){
+                    sh 'flutter config --enable-web'
+                    sh 'pub get'
+                }
+                script{
+                    isTested = false
+                }
+            }
+        }
+        stage('Parallel Stage') {
```
solid-vovabeloded, line 31:
```suggestion
        stage('Run tests') {
```
What do you think?

.jenkins/pipelines/web.groovy
```diff
@@ -0,0 +1,80 @@
...
+                        stage('Run web driver tests') {
+                            steps {
+                                dir('metrics/web'){
+                                    withCredentials([usernamePassword(credentialsId: 'app_credentials', passwordVariable: 'PASSWORD', usernameVariable: 'EMAIL')]) {
+                                        sh 'dart test_driver/main.dart --verbose --store-logs-to=build/logs --email=$EMAIL --password=$PASSWORD --no-verbose'
+                                    }
+                                }
+                            }
+                        }
+                        stage('Run tests') {
```
solid-vovabeloded, line 42:
```suggestion
                        stage('Run flutter tests') {
```
Would it be a bit more clear?

.jenkins/pipelines/web.groovy
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+
+    post {
+        always {
+            cleanWs cleanWhenAborted: true, cleanWhenFailure: true, cleanWhenNotBuilt: true, cleanWhenSuccess: true, cleanWhenUnstable: true, deleteDirs: true, notFailBuild: true
+            build job: 'sync_build_data', propagate: false, wait: false, parameters: [string(name: 'FIREBASE_PROJECT_ID', value: "${env.FIREBASE_PROJECT_ID}"), string(name: 'METRICS_PROJECT_ID', value: "${env.METRICS_PROJECT_ID}"), string(name: 'ANCESTOR_JOB_NAME', value: "${env.JOB_NAME}")]
+        }
+    }
+}
```
solid-vovabeloded, line 80: Would it be great to move the [Monorepo](https://jenkins.gogoapps.dev/job/Monorepo/) pipeline to GitHub as well?

.jenkins/pipelines/ci_integrations.groovy
```diff
@@ -0,0 +1,77 @@
...
+                                }
+                                dir('metrics/ci_integrations'){
+                                    sh 'coverage_converter lcov -i coverage/lcov.info -o coverage-summary.json'
+                                }
+                            }
+                        post {
+                          always {
+                                archiveArtifacts artifacts: 'metrics/ci_integrations/coverage-summary.json', fingerprint: true
+                            }
+                        }
```
solid-vovabeloded, line 64: Would it be better to move this step out of `parallel`? Please, apply this suggestion to the `coverage_converter.groovy` pipeline as well.

.jenkins/pipelines/web.groovy
```diff
@@ -0,0 +1,80 @@
...
+                }
+                dir('metrics/web'){
+                    sh 'coverage_converter lcov -i build/coverage.info -o coverage-summary.json'
+                }
+            }
+            post {
+              always {
+                    archiveArtifacts artifacts: 'metrics/web/coverage-summary.json', fingerprint: true
+                }
+            }
```
solid-vovabeloded, line 70: Would it be better to move this step out of the `Convert coverage report` step?

solid-vovabeloded (changes requested):
Nicely done, @jdzsz! I've added a couple of suggestions here - please take a look. Overall, the pipelines look great!

.jenkins/pipelines/ci_integrations.groovy
```diff
@@ -0,0 +1,66 @@
...
+                dir('metrics/ci_integrations'){
+                    sh 'coverage_converter lcov -i coverage/lcov.info -o coverage-summary.json'
+                }
+            }
+        }
+    }
+
+    post {
+        always {
+            archiveArtifacts artifacts: 'metrics/ci_integrations/coverage-summary.json', fingerprint: true
```
solid-vovabeloded, line 61: I wonder, does the `build job` command will run if the `archiveArtifacts` fail due to the artifact file not found? Maybe we should add `allowEmptyArchive: true` to avoid the situation when the `sync_build_data ` job won't run since we should synchronize the build data even when there is no artifact for some reason (the build failed before running tests, for example).

solid-vovabeloded (changes requested):
Almost there, @jdzsz! I've added one small suggestion here - please take a look.

solid-vovabeloded (approved):
LGTM, thanks @jdzsz!

PR #1049, opened 2021-02-02
feat: buildkite docker pipelines

### LINKED ISSUE #827: Create Docker container to run Buildkite builds
**Description**
To improve Buildkite Runner we want to rely on Docker to make agents less dependent on installed software (flutter,dart).

**Acceptance Criteria**
- [ ] Investigate running Buildkite like builds locally using Docker and build POC

jdzsz:
Fixes #827

### DIFF
--- a/.buildkite/pipelines/ci_integrations_pipeline.yml
+++ b/.buildkite/pipelines/ci_integrations_pipeline.yml
@@ -5,6 +5,11 @@ steps:
       - cd metrics/ci_integrations
       - pub get
       - dartanalyzer . --fatal-infos --fatal-warnings
+    plugins:
+      - docker-compose:
+          config: ".buildkite/compose-files/dart-docker-compose.yml"
+          run: dart
+          workdir: /monorepo
 
   - label: "Run tests"
     key: "tests"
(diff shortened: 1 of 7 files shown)


.buildkite/pipelines/ci_integrations_pipeline.yml
```diff
@@ -13,22 +18,19 @@ steps:
...
-  - label: "Export coverage report"
-    key: "coverage"
-    depends_on: "tests"
-    commands:
-      - curl -o coverage_converter -k https://github.com/platform-platform/monorepo/releases/download/coverage-converter-snapshot/coverage_converter_linux -L
-      - chmod a+x coverage_converter
-      - buildkite-agent artifact download metrics/ci_integrations/coverage/lcov.info .
-      - ./coverage_converter lcov -i metrics/ci_integrations/coverage/lcov.info -o coverage-summary.json
-    artifact_paths: "coverage-summary.json"
+      - coverage_converter lcov -i coverage/lcov.info -o coverage-summary.json
```
solid-vladlaba, line 21: I'm wondering if this step uses the latest release of the `coverage_converter` tool? For example, when we publish a new snapshot for `coverage_converter` it seems like we should build a new version of `testmnrp/dart-converter` to ensure it uses the last release. The same case we have for the `ci_integrations` tool and `testmnrp/sync-builddata`. Do we need to handle this somehow or maybe introduce a new infrastructure to automate updating images with the new tools snapshots?

.buildkite/pipelines/sync_pipeline.yml
```diff
@@ -9,7 +9,17 @@ steps:
     depends_on: "start"
     commands:
       - cd .metrics/buildkite
-      - curl -o ci_integrations -k https://github.com/platform-platform/monorepo/releases/download/ci_integrations-snapshot/ci_integrations_linux -L
-      - chmod a+x ci_integrations
       - eval "echo \"$(sed 's/"/\\"/g' integration_config.yml)\"" >> integration.yml
-      - ./ci_integrations sync --config-file integration.yml
+      - ci_integrations sync --config-file integration.yml
```
solid-vladlaba, line 13: Reminder: please consider a comment for the `coverage_converter`.

.buildkite/pipelines/metrics_web_pipeline.yml
```diff
@@ -4,21 +4,39 @@ steps:
...
+    plugins:
+      - docker-compose:
+          config: ".buildkite/compose-files/flutter-docker-compose.yml"
+          run: flutter
+          workdir: /monorepo
 
   - label: "Flutter test"
     key: "tests"
     depends_on: "analyze"
+    allow_dependency_failure: true
```
solid-vladlaba, line 28: Should we make all pipelines run tests even if analyze step fails? The reason is to make all pipelines similar - if we want to run tests regardless of the `analyze` step result, then we should probably do this for all pipelines.

.buildkite/pipelines/coverage_converter_pipeline.yml
```diff
@@ -13,22 +18,19 @@ steps:
...
-  - label: "Export coverage report"
-    key: "coverage"
-    depends_on: "tests"
-    commands:
-      - curl -o coverage_converter -k https://github.com/platform-platform/monorepo/releases/download/coverage-converter-snapshot/coverage_converter_linux -L
-      - chmod a+x coverage_converter
-      - buildkite-agent artifact download metrics/coverage_converter/coverage/lcov.info .
-      - ./coverage_converter lcov -i metrics/coverage_converter/coverage/lcov.info -o coverage-summary.json
-    artifact_paths: "coverage-summary.json"
+      - coverage_converter lcov -i coverage/lcov.info -o coverage-summary.json
```
solid-vladlaba, line 21: It seems like separating the `tests` and `coverage` steps is more granular and clear though more complex in implementation. The main idea is that coverage exporting failure is not the same as test failure since there can be an error related to the `coverage_converter` tool. Thus, separating these steps allows us to detect the reason for build failure easier and faster.

solid-vladlaba (changes requested):
Well done, @jdzsz! Using Docker within pipelines looks great and more reliable! I've added a couple of suggestions below - please let me know what you think!

illia-romanenko:
NA

PR #1068, opened 2021-02-05
docs: Store database metadata

### LINKED ISSUE #1001: Design store app/db version

solid-vovabeloded:
Fixes #1001 

## Checklist

- [x] Documented


illia-romanenko (approved):

PR #1070, opened 2021-02-06
feat: replace pipelines with dummy ones 

jdzsz:


.buildkite/pipelines/dummy_pipeline.yml
```diff
@@ -0,0 +1,28 @@
+steps:
+  - label: "Dummy step"
+    key: "dummy"
+    commands:
+      - DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get install -y curl 
+      - curl -L https://raw.githubusercontent.com/jdzsz/randomsleep/main/randomsleep.sh -o randomsleep.sh
```
solid-vladlaba, line 6: What if we add this script to the `monorepo` repository so we can maintain it if necessary and there is no need to install `curl` to fetch the script (it would be checked-out with the source code)?

.buildkite/pipelines/dummy_pipeline.yml
```diff
@@ -0,0 +1,28 @@
...
+
+  - label: "Trigger sync for a dummy project"
+    trigger: sync-pipeline
+    depends_on: "finalize"
+    allow_dependency_failure: true
+    async: true
+    build:
+      env:
+        PIPELINE_SLUG: "${BUILDKITE_PIPELINE_SLUG}"
+        METRICS_PROJECT_ID: "buildkite_dummy"
```
solid-vladlaba, line 28: As I can see, there is no project with such ID in Firestore. Should we add it?

.buildkite/pipelines/sync_pipeline.yml
```diff
@@ -9,7 +9,20 @@ steps:
     depends_on: "start"
     commands:
       - cd .metrics/buildkite
+      - DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get install -y curl 
       - curl -o ci_integrations -k https://github.com/platform-platform/monorepo/releases/download/ci_integrations-snapshot/ci_integrations_linux -L
       - chmod a+x ci_integrations
       - eval "echo \"$(sed 's/"/\\"/g' integration_config.yml)\"" >> integration.yml
       - ./ci_integrations sync --config-file integration.yml
```
solid-vladlaba, line 16: Would it be better as we don't generate coverage in the dummy pipeline? So the `ci_integrations` won't try to fetch appropriate artifacts.
```suggestion
      - ./ci_integrations sync --config-file integration.yml --no-coverage
```

solid-vladlaba (changes requested):
Well done, @jdzsz! I've added a couple of suggestions below - please let me know what you think! @solid-vovabeloded will review Jenkins-related parts soon.

.buildkite/pipelines/dummy_pipeline.yml
jdzsz: I forgot to add it, it's added now, same as Jenkins project.

scripts/random_sleep/random_sleep.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+
+TIMEOUT=$(( $RANDOM % 30 ))
+BASE_TIME=$(( ($RANDOM % 59) + 1 ))
+MULTIPLIER=$(( ($RANDOM % 9) + 1 ))
```
solid-vladlaba, line 5: What if we replace `BASE_TIME` and `MULTIPLIER` with a clearer statement in the above? Both cases seem to be very similar mathematically but we introduce one variable instead of two different (we anyway need their product).
```suggestion
EXEC_TIME=$(( $RANDOM % (60*10) + 1 ))
```

If we miss something in the current implementation then we would suggest replacing some numbers as follows: 
```suggestion
BASE_TIME=$(( ($RANDOM % 60) + 1 ))
MULTIPLIER=$(( ($RANDOM % 10) + 1 ))
```
This extends ranges for variables: 
|var|current range|suggested range|
|---|---|---|
|`BASE_TIME`|1..59|1..60|
|`MULTIPLIER`|1..9|1..10|

Credit for discussion @solid-vovabeloded

solid-vladlaba (changes requested):
Almost there, @jdzsz! Two small suggestions here - please review if they make sense! Thanks:+1:

scripts/random_sleep/random_sleep.sh
jdzsz: Actually the output from EXEC_TIME and BASE * MULTIPLIER methods varies a bit. BASE * MULTIPLIER way is created on purpose, it produces mostly short builds and long ones happen rather rarely. EXEC_TIME approach is purely random.  [1000 hits charts comparison](https://imgur.com/KstMFej)

However it is fine for me, I pushed updated script version.

solid-vladlaba: Oh, okay, sorry - now I see what we missed 🤦 
Nevertheless, thank you for the feedback!

solid-vladlaba (approved):
LGTM, thanks @jdzsz!:rocket:

PR #1071, opened 2021-02-08
feat: Publish sha for linux binaries

### LINKED ISSUE #1063: Publish sha for linux binaries of all CLI tools
**Description**
Make sure that we publish sha for linux binaries

**Acceptance Criteria**
- [ ] Publish sha256 file for all linux binary CLI tools

**Additional context**
https://github.com/bazelbuild/rules_kotlin/releases
rules_kotlin_release.tgz.sha256

solid-tetianamovlian:
Fixes #1063

### DIFF
--- a/.github/workflows/ci_integrations_actions.yml
+++ b/.github/workflows/ci_integrations_actions.yml
@@ -86,9 +86,11 @@ jobs:
         working-directory: metrics/ci_integrations/
 
       - name: Generate Linux SHA256 Hash
+        if: matrix.os == 'ubuntu-latest'
         run: shasum metrics/ci_integrations/ci_integrations_linux > metrics/ci_integrations/ci_integrations_linux.sha256
 
       - name: Upload Linux SHA256 Hash
+        if: matrix.os == 'ubuntu-latest'
         uses: actions/upload-artifact@v2
         with:
           name: ci_integrations_linux_sha256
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_integrations_actions.yml
```diff
@@ -91,10 +91,40 @@ jobs:
           name: ${{ matrix.output_name }}
           path: metrics/ci_integrations/${{ matrix.output_name }}
           if-no-files-found: error
+  
+  ci-integrations-linux-generate-sha256-actions:
```
solid-vovabeloded, line 95: I believe it would be better to generate the checksum in the job where we've built the application to avoid the situation when we download the tool with errors and generate the checksum for such a file.

.github/workflows/ci_integrations_actions.yml
```diff
@@ -91,10 +91,40 @@ jobs:
...
+          fetch-depth: 1
+          ref: ${{ github.ref }}
+      
+      - name: Download Linux Build
+        uses: actions/download-artifact@v2
+        with:
+          name: ci_integrations_linux
+          path: metrics/ci_integrations/
+
+      - name: Create CI Integrations Linux Archive
```
solid-vovabeloded, line 112: It seems like we should generate a checksum for a file, but not an archive with this file. Please, apply this suggestion to the `Coverage Converter` checksum generation.

solid-vovabeloded (changes requested):
Well done, @solid-maxim & @solid-tetianamovlian! I've added a couple of suggestions here - please take a look. Overall, the checksum generation looks great, thanks!

solid-vovabeloded (approved):
LGTM, thanks @solid-maxim & @solid-tetianamovlian!

PR #1118, opened 2021-02-19
feat: Add an empty user-agent header for the API calls

### LINKED ISSUE #1105: CLI tools - use empty user-agent
**Description**
We want to add an empty `user-agent` header for the API calls.

**Acceptance Criteria**
- [ ] an empty `user-agent` header is added for the API calls

solid-maxim:
Fixes #1105 

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/metrics/ci_integrations/lib/client/buildkite/buildkite_client.dart
+++ b/metrics/ci_integrations/lib/client/buildkite/buildkite_client.dart
@@ -60,7 +60,7 @@ class BuildkiteClient with LoggerMixin {
   /// A [RegExp] needed to parse next page URLs in [HttpResponse] headers.
   final RegExp _nextUrlRegexp = RegExp('(?<=<)(.*)(?=>)');
 
-  /// A [Map] with HTTP headers, that is added to the default [headers].
+  /// A [Map] with HTTP headers to add to the default [headers] of this client.
   final Map<String, String> _headers;
 
   /// Creates a new instance of the [BuildkiteClient].
--- a/metrics/ci_integrations/lib/client/buildkite/buildkite_client.dart
+++ b/metrics/ci_integrations/lib/client/buildkite/buildkite_client.dart
@@ -1,4 +1,4 @@
-// Use of this source code is governed by the Apache License, Version 2.0 
+// Use of this source code is governed by the Apache License, Version 2.0
 // that can be found in the LICENSE file.
 
 import 'dart:async';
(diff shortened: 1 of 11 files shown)


metrics/ci_integrations/lib/client/buildkite/buildkite_client.dart
```diff
@@ -96,6 +96,7 @@ class BuildkiteClient with LoggerMixin {
     return <String, String>{
       HttpHeaders.contentTypeHeader: ContentType.json.value,
       HttpHeaders.acceptHeader: ContentType.json.value,
+      HttpHeaders.userAgentHeader: null,
```
solid-vladlaba, line 99: What if we introduce the optional `headers` parameter for the client? There also be a private parameter `_headers` with values to apply to all API calls as follows:
```suggestion
      if (_headers != null) ..._headers,
```
In the constructor: `: _headers = headers;`
Thus, we can separate our requirements related to an empty `user-agent` from the low-level client implementation.

solid-vladlaba (changes requested):
Well done, @solid-maxim @solid-tetianamovlian! I've added one suggestion below - please let me know what you think. Also, please apply this suggestion to all of the clients in this PR.

metrics/ci_integrations/lib/source/buildkite/client_factory/buildkite_source_client_factory.dart
```diff
@@ -23,9 +25,13 @@ class BuildkiteSourceClientFactory
     ArgumentError.checkNotNull(config, 'config');
 
     final authorization = BearerAuthorization(config.accessToken);
+    final headers = {
+      HttpHeaders.userAgentHeader: null,
+    };
```
solid-vladlaba, line 30: Would it be better to introduce some class with constants to place there this map with the default headers for the CI Integrations tool? So, here and in other places we can just `SomeConstants.defaultHeaders`. Also, can this map be `const`?
```suggestion
    const headers = {
      HttpHeaders.userAgentHeader: null,
    };
```

metrics/ci_integrations/test/client/buildkite/buildkite_client_test.dart
```diff
@@ -141,6 +141,28 @@ void main() {
...
+          organizationSlug: organizationSlug,
+          authorization: authorization,
+          headers: additionalHeaders,
+        );
+
+        final headers = client.headers;
+
+        final expectedHeader = additionalHeaders.entries.first;
+
+        expect(headers, containsPair(expectedHeader.key, expectedHeader.value));
```
solid-vladlaba, line 162:
```suggestion
        const expectedKey = 'test-header';
        const expectedValue = 'test-value';
        const additionalHeaders = {
          expectedKey: expectedValue,
        };

        final client = BuildkiteClient(
          buildkiteApiUrl: buildkiteMockServer.url,
          organizationSlug: organizationSlug,
          authorization: authorization,
          headers: additionalHeaders,
        );
        final headers = client.headers;
        
        expect(headers, containsPair(expectedKey, expectedValue));
```
Please apply to similar test cases.

solid-vladlaba (dismissed):
Looks better, @solid-maxim @solid-tetianamovlian! A few more small suggestions below - please let me know what you think!

illia-romanenko (approved):
LGTM - thanks @solid-maxim @solid-tetianamovlian !

PR #1119, opened 2021-02-19
docs: CI integrations sync doc

### LINKED ISSUE #1089: ci_integrations: document in-progress sync
**Description**
Create a design document with in-progress sync. Since we don't have proper documentation of the current syn - we should cover other pieces of sync as well.

**Acceptance Criteria**
- [ ] Full algorithm 
- [ ] Describe regular sync process (add Activity diagram if needed)
- [ ] Re-sync process (with Activity diagram)
- [ ] Edge cases (Time-out, history clear, etc...)
- [ ] How it's going to work

solid-vladlaba:
Fixes #1089 

## Checklist

- [x] Documented


metrics/ci_integrations/docs/05_builds_synchronization.md
```diff
@@ -0,0 +1,180 @@
...
+
+Let's consider the following class diagram that contains the main classes and interfaces that participate in the sync process:
+
+![Sync algorithm class diagram](http://www.plantuml.com/plantuml/proxy?cache=no&fmt=svg&src=https://github.com/platform-platform/monorepo/raw/ci_integrations_sync_doc/metrics/ci_integrations/docs/diagrams/sync_algorithm_class_diagram.puml)
+
+When a user starts the synchronization process, the tool invokes the `CiIntegration.sync` method that performs sync algorithm calling `_syncInProgressBuilds` and `_syncBuilds` methods. These methods stand for re-syncing in-progress builds and syncing builds stages respectively. 
+
+The following sequence diagrams detail the two main processes of the sync algorithm:
+
+- **Re-Syncing In-Progress Builds**:
```
illia-romanenko, line 168: I'm trying to understand clearly how timeout is going to be calculated - not sure that I understand it very well. For example what will timeout algorithm do in such case (chronological order):

Build start time b1
Sync time while build in progress s1
Build end time b2
Timeout duration t
Timeout time t1 = b1+t
Sync time s2

b1 --- s1 --- b2 --- t1 --- s2
So if we sync at s2 and during re-sync we receive completed build (failed or successful) so it appears that timeout passed (s2 - b1 > t) but we received build that is completed before timeout duration passed (b2-b1 < t )  (we are not syncing as often) - so are we going to check b2-b1 < t and then decide what to do (set unknown or the real data)? From the other diagrams, I'm assuming that we checking for timeout first right? 

We might want to set clear expectations of what timeout means - is it when (s2 - b1) > t and still in progress? Or something else?

illia-romanenko (changes requested):
Great job @solid-vladlaba - a great description of the new process of re-sync and general sync. I've added a few suggestions/clarifications - please let me know what you think. Detailed diagrams are much appreciated - great job here!

metrics/ci_integrations/docs/05_builds_synchronization.md
solid-vladlaba: As we discussed recently, and as noticed in the document, the build should be timed out if **either**: 
- source failed to load the corresponding build **AND** `dateTimeNow - build.startedAt > timeoutDuration`;
- the corresponding build from the source still in progress **AND** `dateTimeNow - build.startedAt > timeoutDuration`. 

So the `should force timeout` from the activity diagrams of the above sections stands for checking both of these cases.

illia-romanenko (approved):
LGTM - thanks @solid-vladlaba. Excellent work here!

PR #1138, opened 2021-02-26
fix: Add missing closing punctuation to the template document.

solid-vovabeloded:
## Checklist

- [x] Documented


illia-romanenko (approved):
LGTM - thanks @solid-vovabeloded - great catch!

PR #1145, opened 2021-03-01
fix: Remove Validation Delegate abstract class

solid-tetianamovlian:
Fixes #

## Checklist

- [x] Documented
- [x] Unit tested


metrics/ci_integrations/lib/integration/interface/base/config/validation_delegate/validation_delegate.dart
```diff
@@ -1,12 +0,0 @@
-// Use of this source code is governed by the Apache License, Version 2.0 
-// that can be found in the LICENSE file.
-
-import 'package:ci_integration/util/authorization/authorization.dart';
-import 'package:ci_integration/util/model/interaction_result.dart';
-
-/// An abstract class that provides methods for [Config]'s specific
-/// fields validation.
-abstract class ValidationDelegate {
```
solid-vovabeloded, line 9: I believe it would be great to leave a `ValidationDelegate` as a marker interface to create a contract that will say that the `ConfigValidator` should use the `ValidatorDelegate` to validate configuration fields.

solid-vovabeloded (changes requested):
Well done, @solid-tetianamovlian & @solid-maxim! I've added one suggestion here - please take a look.

solid-vovabeloded (approved):

PR #1238, opened 2021-03-26
fix: Fix Firebase Auth Initialization

solid-daniyilyevtyushkin:
Fixes #


solid-vovabeloded (approved):
LGTM, thanks @solid-daniyilyevtyushkin & @solid-yevheniiashchypeleva!

PR #1275, opened 2021-04-09
feat: Extend the BuildkiteClient to include the `fetchBuild` method

### LINKED ISSUE #1274: CI Integrations: Extend the BuildkiteClient to include the `fetchBuild` method
**Description**
Extend the BuildkiteClient to include the `fetchBuild` method.

**Acceptance Criteria**
- [ ] Add `.fetchBuild()` method to the BuildkiteClient.
- [ ] Test new functionality using the BuildkiteMockServer.

solid-daniyilyevtyushkin:
Fixes #1274

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/metrics/ci_integrations/lib/client/buildkite/buildkite_client.dart
+++ b/metrics/ci_integrations/lib/client/buildkite/buildkite_client.dart
@@ -190,6 +190,36 @@ class BuildkiteClient with LoggerMixin {
     );
   }
 
+  /// Fetches a [BuildkiteBuild] by the given [pipelineSlug] and [buildNumber].
+  ///
+  /// Throws an [ArgumentError] if the given [pipelineSlug] or [buildNumber] is
+  /// `null`.
+  Future<InteractionResult<BuildkiteBuild>> fetchBuild(
+    String pipelineSlug,
+    int buildNumber,
+  ) {
+    ArgumentError.checkNotNull(pipelineSlug, 'pipelineSlug');
+    ArgumentError.checkNotNull(buildNumber, 'buildNumber');
+
+    logger.info('Fetching the build #$buildNumber...');
+
+    final url = UrlUtils.buildUrl(
+      basePath,
+      path: 'pipelines/$pipelineSlug/builds/buildNumber',
+    );
+
+    return _handleResponse(
+      _client.get(url, headers: headers),
+      (body, headers) {
+        final buildJson = body as Map<String, dynamic>;
+
+        return InteractionResult.success(
+          result: BuildkiteBuild.fromJson(buildJson),
+        );
+      },
+    );
+  }
+
   /// Fetches a [BuildkiteArtifactPage] by the given [pipelineSlug] and
   /// [buildNumber].
   ///
--- a/metrics/ci_integrations/lib/client/buildkite/buildkite_client.dart
+++ b/metrics/ci_integrations/lib/client/buildkite/buildkite_client.dart
@@ -205,7 +205,7 @@ class BuildkiteClient with LoggerMixin {
 
     final url = UrlUtils.buildUrl(
       basePath,
-      path: 'pipelines/$pipelineSlug/builds/buildNumber',
+      path: 'pipelines/$pipelineSlug/builds/$buildNumber',
     );
 
     return _handleResponse(
--- a/metrics/ci_integrations/lib/client/buildkite/buildkite_client.dart
+++ b/metrics/ci_integrations/lib/client/buildkite/buildkite_client.dart
@@ -190,16 +190,17 @@ class BuildkiteClient with LoggerMixin {
     );
   }
 
-  /// Fetches a [BuildkiteBuild] by the given [pipelineSlug] and [buildNumber].
+  /// Fetches a [BuildkiteBuild] of a pipeline with the given [pipelineSlug] 
+  /// and having the given [buildNumber].
   ///
-  /// Throws an [ArgumentError] if the given [pipelineSlug] or [buildNumber] is
+  /// Throws an [AssertionError] if the given [pipelineSlug] or [buildNumber] is
   /// `null`.
   Future<InteractionResult<BuildkiteBuild>> fetchBuild(
     String pipelineSlug,
     int buildNumber,
   ) {
-    ArgumentError.checkNotNull(pipelineSlug, 'pipelineSlug');
-    ArgumentError.checkNotNull(buildNumber, 'buildNumber');
+    assert(pipelineSlug != null);
+    assert(buildNumber != null);
 
     logger.info('Fetching the build #$buildNumber...');
 
--- a/metrics/ci_integrations/test/client/buildkite/buildkite_client_test.dart
+++ b/metrics/ci_integrations/test/client/buildkite/buildkite_client_test.dart
@@ -20,6 +20,7 @@ void main() {
   group("BuildkiteClient", () {
     const testPageNumber = 1;
     const buildNumber = 1;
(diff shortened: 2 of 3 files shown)


metrics/ci_integrations/lib/client/buildkite/buildkite_client.dart
```diff
@@ -190,6 +190,36 @@ class BuildkiteClient with LoggerMixin {
...
+  /// Fetches a [BuildkiteBuild] by the given [pipelineSlug] and [buildNumber].
+  ///
+  /// Throws an [ArgumentError] if the given [pipelineSlug] or [buildNumber] is
+  /// `null`.
+  Future<InteractionResult<BuildkiteBuild>> fetchBuild(
+    String pipelineSlug,
+    int buildNumber,
+  ) {
+    ArgumentError.checkNotNull(pipelineSlug, 'pipelineSlug');
+    ArgumentError.checkNotNull(buildNumber, 'buildNumber');
```
solid-vladlaba, line 202: I'm not sure we should check this here as we don't do this in most of the methods of the client. Moreover, if we want to make this validation more developer-oriented then wouldn't it be better to replace error throwing with asserts? For production and usage purposes, I suppose it's much better when the client returns InteractionResult - especially for adapters, so each client call won't be wrapped with a try-catch. There are few points: 
- usually, the server provides more detailed errors with descriptions;
- when throwing additional errors instead of returning InteractionResult, one should consider providing an error handling to make errors clear to end-users;
- `asserts` more dev-oriented as don't work in production;
- in other places, where we're throwing `ArgumentError`, we mean that the process should be interrupted as the invalid argument makes it impossible to proceed with an algorithm - here, we still can make an API call: InteractionResult.error will be returned and the algorithm will decide what to do next.

So, according to the above, I would suggest to replace errors with asserts:
```suggestion
    assert(pipelineSlug != null);
    assert(buildNumber != null);
```
or completely remove these lines:
```suggestion
```

metrics/ci_integrations/lib/client/buildkite/buildkite_client.dart
```diff
@@ -190,6 +190,36 @@ class BuildkiteClient with LoggerMixin {
     );
   }
 
+  /// Fetches a [BuildkiteBuild] by the given [pipelineSlug] and [buildNumber].
```
solid-vladlaba, line 193:
```suggestion
  /// Fetches a [BuildkiteBuild] of a pipeline with the given [pipelineSlug] and having the given [buildNumber].
```

metrics/ci_integrations/test/client/buildkite/buildkite_client_test.dart
```diff
@@ -398,10 +399,66 @@ void main() {
       },
     );
 
+    test(
+      ".fetchBuild() throws an ArgumentError if the given pipeline slug is null",
+      () async {
```
solid-vladlaba, line 404:
```suggestion
      () {
```
?

metrics/ci_integrations/test/client/buildkite/buildkite_client_test.dart
```diff
@@ -398,10 +399,66 @@ void main() {
...
+        expect(() => client.fetchBuild(null, buildNumber), throwsArgumentError);
```
solid-vladlaba, line 405:
```suggestion
        final result = client.fetchBuild(null, buildNumber);
        expect(result, throwsArgumentError);
```
Would it work as the `result` is the`Future`? If yes, then this seems to be more readable.

metrics/ci_integrations/test/client/buildkite/buildkite_client_test.dart
```diff
@@ -398,10 +399,66 @@ void main() {
...
+      },
+    );
+
+    test(
+      ".fetchBuild() throws an ArgumentError if the given build number is null",
```
solid-vladlaba, line 410: Please apply similar suggestions as the above ones.

metrics/ci_integrations/test/client/buildkite/buildkite_client_test.dart
```diff
@@ -398,10 +399,66 @@ void main() {
...
+          notFound,
+          buildNumber,
+        );
+
+        expect(interactionResult.isError, isTrue);
+      },
+    );
+
+    test(
+      ".fetchBuild() returns an error if the build with the build with the given build number is not found",
```
solid-vladlaba, line 432:
```suggestion
      ".fetchBuild() returns an error if a build with the given build number is not found",
```

metrics/ci_integrations/test/client/buildkite/buildkite_client_test.dart
```diff
@@ -398,10 +399,66 @@ void main() {
...
+      () async {
+        expect(
+          () => client.fetchBuild(pipelineSlug, null),
+          throwsArgumentError,
+        );
+      },
+    );
+
+    test(
+      ".fetchBuild() returns an error if the pipeline with the given slug is not found",
```
solid-vladlaba, line 420:
```suggestion
      ".fetchBuild() returns an error if a pipeline with the given pipeline slug is not found",
```

metrics/ci_integrations/test/client/buildkite/buildkite_client_test.dart
```diff
@@ -398,10 +399,66 @@ void main() {
...
+
+    test(
+      ".fetchBuild() returns a buildkite build",
+      () async {
+        final interactionResult = await client.fetchBuild(
+          pipelineSlug,
+          buildNumber,
+        );
+
+        expect(interactionResult.result, isNotNull);
```
solid-vladlaba, line 451: Let's also check that the interaction is success here to ensure an algorithm is correct: 
```suggestion
        expect(interactionResult.isSuccess, isTrue);
        expect(interactionResult.result, isNotNull);
```

Should we also check that result has the number as we requested?

metrics/ci_integrations/test/client/buildkite/test_utils/mock/buildkite_mock_server.dart
```diff
@@ -126,6 +144,15 @@ class BuildkiteMockServer extends ApiMockServer {
     await MockServerUtils.writeResponse(request, body: response);
   }
 
+  /// Responses with a [BuildkiteBuild].
+  Future<void> _buildResponse(HttpRequest request) async {
+    const build = BuildkiteBuild();
+
+    final response = build.toJson();
```
solid-vladlaba, line 151:
```suggestion
    final data = build.toJson();
```

metrics/ci_integrations/test/client/buildkite/test_utils/mock/buildkite_mock_server.dart
```diff
@@ -126,6 +144,15 @@ class BuildkiteMockServer extends ApiMockServer {
     await MockServerUtils.writeResponse(request, body: response);
   }
 
+  /// Responses with a [BuildkiteBuild].
```
solid-vladlaba, line 147: Should we also set the correct build number as requested (provided in `request`)?

solid-vladlaba (changes requested):
Well done, @solid-daniyilyevtyushkin @solid-yevheniiashchypeleva! I've added a couple of suggestions below - please let me know what you think! Overall, the code looks good and well-tested, thanks!

solid-vladlaba (approved):
LGTM, thanks @solid-daniyilyevtyushkin @solid-yevheniiashchypeleva!:rocket:

PR #1338, opened 2021-04-27
feat: Builds aggregation - Domain Layer

### LINKED ISSUE #1331: [Web] Builds Aggregation - Domain layer
**Description**
Implement the Builds Aggregation domain layer according to the design.

**Acceptance Criteria**
- [ ] The `BuildDayRepository` interface implemented.
- [ ] The `BuildDay` entity created.
- [ ] The `ReceiveBuildDayProjectMetricsUpdates` class created.
- [ ] The `BuildDayProjectMetrics` class created.
- [ ] All classes are properly tested.

**Additional context**
[Builds Aggregation design](https://github.com/platform-platform/monorepo/blob/7d1a706eb5f01092ccd76790bc0fbbc357a55cf2/metrics/web/docs/features/builds_aggregation/01_builds_aggregation_design.md#domain-layer)

solid-daniyilyevtyushkin:
Fixes #1331

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/metrics/web/test/dashboard/domain/entities/metrics/receive_build_day_project_metrics_updates_test.dart
+++ b/metrics/web/test/dashboard/domain/entities/metrics/receive_build_day_project_metrics_updates_test.dart
@@ -1,10 +1,12 @@
 // Use of this source code is governed by the Apache License, Version 2.0
 // that can be found in the LICENSE file.
 
+import 'package:collection/collection.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:metrics/common/domain/entities/build_day.dart';
 import 'package:metrics/common/domain/repositories/build_day_repository.dart';
 import 'package:metrics/dashboard/domain/entities/collections/date_time_set.dart';
+import 'package:metrics/dashboard/domain/entities/metrics/build_day_project_metrics.dart';
 import 'package:metrics/dashboard/domain/entities/metrics/build_performance.dart';
 import 'package:metrics/dashboard/domain/usecases/parameters/project_id_param.dart';
 import 'package:metrics/dashboard/domain/usecases/receive_build_day_project_metrics_updates.dart';
@@ -16,11 +18,12 @@ void main() {
   group("ReceiveBuildDayProjectMetricsUpdates", () {
     const projectId = 'id';
     const projectIdParam = ProjectIdParam(projectId);
+    const collectionEquality = DeepCollectionEquality();
 
     final buildDayRepository = _BuildDayRepositoryMock();
     final useCase = ReceiveBuildDayProjectMetricsUpdates(buildDayRepository);
 
-    PostExpectation<Stream<List<BuildDay>>> whenFetchBuildDaysStream() {
+    PostExpectation<Stream<List<BuildDay>>> whenStreamBuildDays() {
       return when(buildDayRepository.projectBuildDaysInDateRangeStream(
         projectId,
         from: anyNamed('from'),
@@ -46,20 +49,16 @@ void main() {
       );
     }
 
-    int calculateTotalNumberOfBuilds(List<BuildDay> buildDays) {
-      return buildDays.fold<int>(0, (value, buildDay) {
-        return value +
-            buildDay.successful +
-            buildDay.failed +
-            buildDay.unknown +
-            buildDay.inProgress;
-      });
-    }
-
-    Duration calculateTotalBuildsDuration(List<BuildDay> buildDays) {
-      return buildDays.fold<Duration>(Duration.zero, (value, buildDay) {
-        return value + buildDay.totalDuration;
-      });
+    Matcher buildDayMetricsMatcher({
+      Matcher projectIdMatcher = anything,
+      Matcher buildNumberMetricMatcher = anything,
+      Matcher performanceMetricMatcher = anything,
+    }) {
+      return allOf(
+        projectIdMatcher,
+        buildNumberMetricMatcher,
+        performanceMetricMatcher,
+      );
     }
 
     tearDown(() {
@@ -87,19 +86,9 @@ void main() {
     );
 
     test(
-      "successfully creates an instance with the given build day repository",
-      () {
-        expect(
-          () => ReceiveBuildDayProjectMetricsUpdates(buildDayRepository),
-          returnsNormally,
-        );
-      },
-    );
-
-    test(
       ".call() delegates to the .projectBuildDaysInDateRangeStream() with the given project id and a non-null from date",
       () {
-        whenFetchBuildDaysStream().thenAnswer((_) => Stream.value([]));
+        whenStreamBuildDays().thenAnswer((_) => Stream.value([]));
 
         useCase.call(projectIdParam);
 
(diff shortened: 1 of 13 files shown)


metrics/web/test/dashboard/domain/entities/metrics/receive_build_day_project_metrics_updates_test.dart
```diff
@@ -0,0 +1,263 @@
...
+      },
+    );
+
+    test(
+      ".call() returns a build day project metrics with the given project id",
+      () async {
+        whenFetchBuildDaysStream().thenAnswer((_) => Stream.value([]));
+
+        final metricsStream = useCase.call(projectIdParam);
+        final buildDayMetrics = await metricsStream.firstWhere(
```
solid-vladlaba, line 119: Should we maybe use the [emits](https://pub.dev/documentation/test_api/latest/expect/emits.html) matcher where possible that is purposed to test streams?

metrics/web/test/dashboard/domain/entities/metrics/receive_build_day_project_metrics_updates_test.dart
```diff
@@ -0,0 +1,263 @@
...
+      },
+    );
+
+    test(
+      ".call() returns a build day project metrics with the build number metric equal to the total number of builds performed during the returned build days",
+      () async {
+        final firstBuildDay = createBuildDay(successful: 1, failed: 1);
+        final secondBuildDay = createBuildDay(unknown: 2, inProgress: 2);
+        final buildDays = [firstBuildDay, secondBuildDay];
+        final expectedNumberOfBuilds = calculateTotalNumberOfBuilds(buildDays);
```
solid-vladlaba, line 147: The `calculateTotalNumberOfBuilds` function here is a potential item for testing by itself. Moreover, it is implemented in the same way as builds are calculated within the use case under tests. This makes tests confusing as they know how to calculate everything that is calculated inside the use case. Thus, we should either move a logic with the required calculation elsewhere and test it separately or use concrete values in tests. Here, we definitely know that the number of builds for the specified list is `1 + 1 + 2 + 2 = 6`. As we use **concrete** test data, we expect **concrete** results and we know them. The same idea is applied to the `calculateTotalBuildsDuration `.
```suggestion
        const expectedNumberOfBuilds = 1 + 1 + 2 + 2;
```
or, which is the same:
```suggestion
        const expectedNumberOfBuilds = 6;
```

metrics/web/test/dashboard/domain/entities/metrics/receive_build_day_project_metrics_updates_test.dart
```diff
@@ -0,0 +1,263 @@
...
+        final buildDayMetrics = await metricsStream.firstWhere(
+          (metrics) => metrics != null,
+        );
+
+        expect(buildDayMetrics.performanceMetric, isNotNull);
+      },
+    );
+
+    test(
+      ".call() returns a build day project metrics with the performance metric having zero average build duration if the returned build days don't contain any builds",
```
solid-vladlaba, line 177:
```suggestion
      ".call() returns a build day project metrics with the performance metric having zero average build duration if the returned build days have no builds",
```

solid-vladlaba (changes requested):
Well done, @solid-daniyilyevtyushkin @solid-yevheniiashchypeleva! I've added a couple of suggestions below - please let me know what you think!

metrics/web/test/dashboard/domain/entities/metrics/receive_build_day_project_metrics_updates_test.dart
```diff
@@ -0,0 +1,308 @@
...
+import 'package:metrics/dashboard/domain/entities/metrics/build_day_project_metrics.dart';
+import 'package:metrics/dashboard/domain/entities/metrics/build_performance.dart';
+import 'package:metrics/dashboard/domain/usecases/parameters/project_id_param.dart';
+import 'package:metrics/dashboard/domain/usecases/receive_build_day_project_metrics_updates.dart';
+import 'package:mockito/mockito.dart';
+
+import '../../../../test_utils/matchers.dart';
+
+void main() {
+  group("ReceiveBuildDayProjectMetricsUpdates", () {
```
solid-vladlaba, line 18: It seems like this file is mislocated. Should we place it under the `test/dashboard/domain/usecases`?

metrics/web/test/dashboard/domain/entities/metrics/receive_build_day_project_metrics_updates_test.dart
```diff
@@ -0,0 +1,308 @@
...
+    Matcher buildDayMetricsMatcher({
+      Matcher projectIdMatcher = anything,
+      Matcher buildNumberMetricMatcher = anything,
+      Matcher performanceMetricMatcher = anything,
+    }) {
+      return allOf(
+        projectIdMatcher,
+        buildNumberMetricMatcher,
+        performanceMetricMatcher,
+      );
```
solid-vladlaba, line 61: Great idea of using the `allOf` matcher 👍 
But what if we'll try to simplify using this `buildDayMetricsMatcher`. Currently, we have to pass a `predicate` that explicitly accesses the concrete value of the `BuildDayProjectMetrics`. How about using the `isA` matcher with the `having` construction as follows:
 ```suggestion
      final metricsMatcher = isA<BuildDayProjectMetrics>();
      return allOf([
        metricsMatcher.having(
          (metrics) => metrics.projectId,
          'projectId',
          projectIdMatcher,
        ),
        metricsMatcher.having(
          (metrics) => metrics.buildNumberMetric,
          'buildNumberMetric',
          buildNumberMetricMatcher,
        ),
        metricsMatcher.having(
          (metrics) => metrics.performanceMetric,
          'performanceMetric',
          performanceMetricMatcher,
        ),
      ]);
```
This allows passing matches for concrete fields and their values instead of using a predicate every time. For example, in the appropriate test cases, instead of: 
```dart
        final projectIdMatcher = predicate<BuildDayProjectMetrics>(
          (metrics) => metrics.projectId == expectedProjectId,
        );

        expect(
          metricsStream,
          emits(buildDayMetricsMatcher(projectIdMatcher: projectIdMatcher)),
        );
```
we can write: 
```dart
        final metricsMatcher = buildDayMetricsMatcher(
          projectIdMatcher: equals(expectedProjectId),
        );

        expect(metricsStream, emits(metricsMatcher));
```

What do you think?

metrics/web/test/dashboard/domain/entities/metrics/receive_build_day_project_metrics_updates_test.dart
```diff
@@ -0,0 +1,308 @@
...
+          successful: 2,
+        );
+        final secondBuildDay = createBuildDay(
+          totalDuration: secondDayDuration,
+          failed: 3,
+        );
+        final buildDays = [firstBuildDay, secondBuildDay];
+        final totalDuration = firstDayDuration + secondDayDuration;
+        const totalNumberOfBuilds = 2 + 3;
+        final expectedAverageDuration = totalDuration ~/ totalNumberOfBuilds;
```
solid-vladlaba, line 235:
```suggestion
        final expectedAverageDuration = Duration(seconds: 1);
```

metrics/web/test/dashboard/domain/entities/metrics/receive_build_day_project_metrics_updates_test.dart
```diff
@@ -0,0 +1,308 @@
...
+            return averageDuration == expectedAverageDuration;
+          },
+        );
+
+        expect(
+          metricsStream,
+          emits(buildDayMetricsMatcher(
+            performanceMetricMatcher: performanceMetricMatcher,
+          )),
+        );
```
solid-vladlaba, line 255: Example.
If apply the suggestion from the above (with metrics matcher), we can write as follows:
```suggestion
        final metricsMatcher = buildDayMetricsMatcher(
          performanceMetricMatcher: isA<PerformanceMetric>().having(
            (metric) => metric.averageBuildDuration,
            'averageBuildDuration',
            equals(expectedAverageDuration),
          ),
        );

        expect(metricsStream, emits(metricsMatcher));
```

solid-vladlaba (changes requested):
Almost there, @solid-daniyilyevtyushkin @solid-yevheniiashchypeleva! I've added some small suggestions below - please let me know what you think!

solid-vladlaba (approved):
LGTM, thanks @solid-daniyilyevtyushkin @solid-yevheniiashchypeleva!:rocket:

PR #1349, opened 2021-04-30
docs: Update concept map to include all Metrics components

### LINKED ISSUE #1347: Update concept map to include cloud functions and metrics CLI
**Description**
Update concept map to include cloud functions (green, dart) and metrics CLI as a standalone installer component.

**Acceptance Criteria**
- [ ] Concept map now includes cloud functions and metrics CLI.

**Additional context**
[Concept map](https://github.com/platform-platform/monorepo/blob/master/concept_map.puml)

solid-vovabeloded:
Fixes #1347

## Checklist

- [x] Documented

### DIFF
--- a/concept_map.puml
+++ b/concept_map.puml
@@ -40,13 +40,13 @@ together {
 
     rectangle Core
 }
-rectangle Web
+
+rectangle "Flutter Web" as Web
 rectangle "Deploy CLI" as CLI
 
 Jenkins -[hidden]right-> GA
 GA -[hidden]right-> Buildkite
-
-Core -[hidden]left-> Analytics  
+Analytics -[hidden]left-> Firestore  
 
 Jenkins -down-> CI : data
 GA -down-> CI : data
@@ -54,23 +54,20 @@ Buildkite -down-> CI : data
 
 CI -down-> Firestore : sync data
 CI -down-> Authentication : authenticates
-CI -[hidden]down-> Hosting
 
 Hosting -[hidden]right-> Authentication
 Authentication -[hidden]right-> Firestore
 
-Web --up--> Firestore : reads metrics and \l stores app config
-Web --up--> Authentication : authenticates
-Web --up--> Hosting : deployed on
-Web --up--> Analytics : reports analytics data
-Web -down-> Sentry : logging
-
-DartCloudFunctions -[hidden]-> Core : uses
+Web -up-> Firestore : reads metrics and \l stores app config
+Web -up-> Authentication : authenticates
+Web -up-> Hosting : deployed on
+Web .up.> Analytics : reports analytics data \l (optional)
+Web .down.> Sentry : logging \l (optional)
 
 CLI -up-> Core : uses
 
-DartCloudFunctions -down-> Core : uses
-Web --right--> Core : uses
-CI --right--> Core  : uses
+DartCloudFunctions -right-> Core : uses
+Web -right-> Core : uses
+CI -right-> Core  : uses
 
 @enduml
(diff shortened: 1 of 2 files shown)


concept_map.puml
```diff
@@ -26,22 +26,29 @@ cloud "GitHub Actions" as GA
...
 rectangle "CI integrations" as CI
+rectangle "Dart Cloud Functions" as DartCloudFunctions
 
 together {
     node Firebase {
         cloud Hosting
         database Firestore
         cloud Authentication
+        cloud "Cloud Functions" as CloudFunctions
+        cloud Analytics
```
solid-vladlaba, line 37: What do you think if we reduce some "clouds" from the concept map? For example, by making CIs in the above and Sentry as `node` or something else.

concept_map.puml
```diff
@@ -26,22 +26,29 @@ cloud "GitHub Actions" as GA
...
 
 rectangle "CI integrations" as CI
+rectangle "Dart Cloud Functions" as DartCloudFunctions
 
 together {
     node Firebase {
         cloud Hosting
         database Firestore
         cloud Authentication
+        cloud "Cloud Functions" as CloudFunctions
```
solid-vladlaba, line 36: Generally speaking, the Cloud Functions and Dart Cloud Functions are the same entity for this diagram. What if we inline it or notice somehow that our functions are written in Dart?
```suggestion
        cloud "Dart Cloud Functions" as DartCloudFunctions
```
We can make this cloud green noticing that this is part of Metrics components.

concept_map.puml
```diff
@@ -53,12 +60,20 @@ CI -[hidden]down-> Hosting
 Hosting -[hidden]right-> Authentication
 Authentication -[hidden]right-> Firestore
 
-Firebase -[hidden]right-> Core
+Web --up--> Firestore : reads metrics and \l stores app config
+Web --up--> Authentication : authenticates
+Web --up--> Hosting : deployed on
+Web --up--> Analytics : reports analytics data
+Web -down-> Sentry : reports about errors
```
solid-vladlaba, line 67:
```suggestion
Web -down-> Sentry : logging
```
Actually, Sentry may be used not for errors only so maybe we can write logging here.

concept_map.puml
```diff
@@ -53,12 +60,20 @@ CI -[hidden]down-> Hosting
...
+
+DartCloudFunctions -up-> CloudFunctions : deployed on
+DartCloudFunctions -[hidden]-> Core : uses
+DartCloudFunctions -up-> Firestore : creates aggregation \l documents
```
solid-vladlaba, line 71: Consider the above suggestion with `CloudFunctions`. Also, it seems like the aggregation message reveals some unnecessary details here. So, how about that?
```suggestion
DartCloudFunctions -[hidden]-> Core : uses
```

concept_map.puml
```diff
@@ -53,12 +60,20 @@ CI -[hidden]down-> Hosting
...
 
-Web -up-> Firestore : reads metrics and \l stores app config
-Web -up-> Authentication : authenticates
-Web -up-> Hosting : deployed on
+CLI --> Web : deploys
```
solid-vladlaba, line 73: Not sure the CLI deploys the Web only. Instead, what if we rename CLI into `Deploy CLI` on the diagram? Then, keep the `CLI --> Core : uses` but remove the `deploys` line:
```suggestion
```
What do you think?

solid-vladlaba (commented):
Well done, @solid-vovabeloded! I've added a couple of suggestions below - please let me know what you think! The new concept map diagram looks great 👍

illia-romanenko (approved):
LGTM - thanks @solid-vovabeloded and @solid-vladlaba. Thanks for all of the work you put to polish it! Also, FYI created https://github.com/platform-platform/monorepo/issues/1354 to improve it even further.

PR #1357, opened 2021-05-05
docs: Update Firebase builds aggregation document

solid-tetianamovlian:
Fixes #

## Checklist

- [x] Documented


metrics/firebase/docs/features/builds_aggregation/diagrams/firestore_create_builds_aggregation_sequence_diagram.puml
```diff
@@ -17,6 +17,12 @@
     FCF -> FCF : triggers an onCreate event handler
     
     FCF -> Firestore : increases a counter inside the 'build_days' collection
+    FCF -> Firestore : increases a successful builds duration inside the 'build_days' collection
```
solid-vovabeloded, line 20: It seems like we should increment the `successfulBuildsduration` only in case when the build status is successful. So, I believe we should show on this diagram that we are incrementing the `counter` inside the document in any case and then explain that we are incrementing the `successfulBuildsduration` in case the created build has the `successful` build status. The same suggestion for the `update builds aggregation` sequence diagram.

metrics/firebase/docs/features/builds_aggregation/01_firestore_builds_aggregation_design.md
```diff
@@ -180,4 +182,4 @@ The `onCreate` handler takes two arguments - `DocumentSnapshot` and `EventContex
 
 The `onUpdate` event handler, has a similar second argument, but the first one is a `Change`. As it has two states: after the update event and prior to the event, we should create an instance of `Change` class using the mocked `DocumentSnapshot`s and pass it as the first argument to the handler.
 
-As our functions process some actions(creating documents or updating existing ones) using the Firestore instance from the `DocumentSnapshot` through `documentSnapshot.firestore` we can use our mocked instance to override this call and replace it with our created mock. This gives us the ability to verify that the functions actually write or updates documents.
+As our functions process some actions (creating documents or updating existing ones) using the Firestore instance from the `DocumentSnapshot` through `documentSnapshot.firestore` we can use our mocked instance to override this call and replace it with our created mock. This gives us the ability to verify that the functions actually write or updates documents.
```
solid-vovabeloded, line 185:
```suggestion
As our functions process some actions (creating documents or updating existing ones) using the Firestore instance from the `DocumentSnapshot` through `documentSnapshot.firestore`, we can use our mocked instance to override this call and replace it with our created mock. This gives us the ability to verify that the functions actually write or updates documents.
```

solid-vovabeloded (changes requested):
Great job, @solid-maxim & @solid-tetianamovlian! I've added a couple of suggestions here - please take a look. Overall, the document updates look nice - thanks!

solid-vovabeloded (approved):

PR #1380, opened 2021-05-17
feat: Deployer prompt improvements

solid-daniyilyevtyushkin:
Fixes #

## Checklist
- [x] Unit tested

### DIFF
--- a/metrics/cli/lib/deploy/deployer.dart
+++ b/metrics/cli/lib/deploy/deployer.dart
@@ -130,7 +130,10 @@ class Deployer {
         deployPaths.firebasePath,
         deployPaths.webAppPath,
       );
+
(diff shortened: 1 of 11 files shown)


solid-vladlaba (changes requested):
Well done, @solid-daniyilyevtyushkin @solid-yevheniiashchypeleva! I've added one small suggestion below - please let me know what you think! Overall, the changes look great, however, I suppose we should rethink the prompter and writers a bit in the future. It seems like usual info text differs a lot from other prompter's responsibilities and can be confusing. But let's keep it as is for now.

solid-vladlaba (approved):
LGTM, thanks @solid-daniyilyevtyushkin @solid-yevheniiashchypeleva !:rocket:

PR #1412, opened 2021-05-24
fix: Fix Sentry enter release name string

solid-tetianamovlian:
Fixes #


solid-vovabeloded (approved):

PR #1461, opened 2021-06-16
fix: Add dependency override

solid-yevheniiashchypeleva:
Fixes #


solid-vladlaba (approved):
LGTM, thanks!

@illia-romanenko as this is a temporary solution, I suppose we should consider upgrading the Flutter version for the project.

PR #1477, opened 2021-07-05
feat: Add an ability to set the auth token for the whole CLI class

### LINKED ISSUE #1474: [Metrics CLI] Add an ability to set the auth token for the whole CLI class
**Description**
Implement an abstract class for the CLIs that use the auth token for their commands. The abstract class should have the ability to set the `auth token`. 

**Acceptance Criteria**
- [ ] Implement an abstract CLI class that is responsible for working with the `auth token`
- [ ] CLIs that work with the auth token should inherit this abstract class

solid-dimakoniaiev:
Fixes #1474 

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/metrics/cli/lib/services/common/mixin/authorization.dart
+++ b/metrics/cli/lib/services/common/mixin/authorization.dart
@@ -1,0 +1,11 @@
+// Use of this source code is governed by the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+/// A mixin that provides common methods for setting the authorization up.
+mixin Authorization {
+  /// Initializes the given [authorization] for this service.
+  void initializeAuthorization(String authorization);
+
+  /// Resets this service current authorization.
+  void resetAuthorization();
+}
--- a/metrics/cli/test/services/common/auth_cli_test.dart
+++ b/metrics/cli/test/services/common/auth_cli_test.dart
@@ -8,16 +8,16 @@ import 'package:test/test.dart';
 
 void main() {
   group("AuthCli", () {
-    const authorization = 'authorization';
+    const auth = 'auth';
 
     final testArguments = ['testArgument1', 'testArgument2'];
 
-    String composeAuthArgument(String authArgumentName, String authorization) {
-      return '--$authArgumentName=$authorization';
+    String composeAuthArgument(String authArgumentName, String authValue) {
+      return '--$authArgumentName=$authValue';
     }
 
     test(
-      ".setupAuth() throws an ArgumentError if the given authorization is null",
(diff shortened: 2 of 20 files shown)


metrics/cli/lib/services/common/mixin/authorization.dart
```diff
@@ -0,0 +1,11 @@
+// Use of this source code is governed by the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+/// A mixin that provides common methods for setting the authorization up.
+mixin Authorization {
```
solid-vladlaba, line 5: 🏗️ 
Not sure this should be a mixin as mixins are most likely used to share code between classes. In our case, we can pass the auth to adapters so they can explicitly use the given value where necessary. Otherwise, we should think about creating the `AuthService` interface as we have for the `InfoService`.

metrics/cli/test/services/common/auth_cli_test.dart
```diff
@@ -0,0 +1,38 @@
+// Use of this source code is governed by the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+import 'package:cli/services/common/cli/auth_cli.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group("AuthCli", () {
+    test(
```
solid-vladlaba, line 9: ✅ 
Should we also test that the auth cli adds the specified auth argument to the beginning or end of the arguments list? Also, that it composes argument and value correctly?

solid-vladlaba (changes requested):
Well done, @solid-dimakoniaiev! I've added a couple of suggestions below - please let me know what you think!

metrics/cli/test/services/common/auth_cli_test.dart
```diff
@@ -0,0 +1,139 @@
...
+        final expected = composeAuthArgument(authArgumentName, authorization);
+
+        await authCli.runWithAuth(testArguments);
+
+        expect(authCli.arguments.last, expected);
+      },
+    );
+
+    test(
+      ".runWithAuth() does nothing with arguments list if the authorization is null",
```
solid-vladlaba, line 82: 🏷️ 
```suggestion
      ".runWithAuth() does not add authorization to arguments if the authorization value is null",
```

solid-vladlaba (changes requested):
Almost there, @solid-dimakoniaiev! I've added a couple of suggestions below - please let me know what you think!

solid-vladlaba (approved):
LGTM, thanks @solid-dimakoniaiev!:rocket:

PR #1494, opened 2021-07-19
feat: Implement Update command

### LINKED ISSUE #1466: [Metrics CLI] Implement UpdateCommand
**Description**
Implement UpdateCommand according to the Update feature design and integrate it into the Metrics CLI

**Acceptance Criteria**
- [ ] The UpdateCommand implemented
- [ ] The UpdateCommand integrated to the Metrics CLI

**Additional context**
[UpdateCommand](https://github.com/Flank/flank-dashboard/blob/d2f6d1f5f5e7df4d5ea30ff7c9f933d4f3f3e2d9/metrics/cli/docs/features/update/06_update_feature.md#updatecommand)

solid-dimakoniaiev:
Fixes #1466 

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/metrics/cli/lib/cli/command/update_command.dart
+++ b/metrics/cli/lib/cli/command/update_command.dart
@@ -1,0 +1,55 @@
+// Use of this source code is governed by the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+import 'package:args/command_runner.dart';
+import 'package:cli/cli/updater/config/factory/update_config_factory.dart';
+import 'package:cli/cli/updater/factory/updater_factory.dart';
+import 'package:cli/cli/updater/updater.dart';
+import 'package:cli/common/model/config/update_config.dart';
+
+/// A [Command] that updates the deployed Metrics components.
+class UpdateCommand extends Command {
+  /// A name of the option that holds a path to the YAML configuration file.
+  static const _configFileOptionName = 'config-file';
+
+  @override
+  final name = 'update';
+
+  @override
+  final description = 'Updates the deployed Metrics components.';
+
+  /// An [UpdaterFactory] this command uses to create a [Updater].
+  final UpdaterFactory _updaterFactory;
+
+  /// An [UpdateConfigFactory] this command uses to create an [UpdateConfig].
+  final UpdateConfigFactory _updateConfigFactory;
+
+  /// Creates a new instance of the [UpdateCommand].
+  ///
+  /// If the given [updateConfigFactory] is `null`, an instance of the
+  /// [UpdateConfigFactory] is used.
+  ///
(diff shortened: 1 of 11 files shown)


metrics/cli/lib/cli/command/update_command.dart
```diff
@@ -0,0 +1,55 @@
...
+  /// A name of the option that holds a path to the YAML configuration file.
+  static const _configFileOptionName = 'config-file';
+
+  @override
+  final name = 'update';
+
+  @override
+  final description = 'Updates the deployed Metrics components.';
+
+  /// An [UpdaterFactory] this command uses to create a [Updater].
```
solid-vladlaba, line 21: 🏷️ 
```suggestion
  /// An [UpdaterFactory] this command uses to create an [Updater].
```

metrics/cli/test/cli/command/update_command_test.dart
```diff
@@ -0,0 +1,189 @@
...
+
+        verify(updater.update(updateConfig)).called(once);
+      },
+    );
+  });
+}
+
+/// A stub class for an [UpdateCommand] class providing test implementation
+/// for methods.
+class _UpdateCommandStub extends UpdateCommand {
```
solid-vladlaba, line 136: ✅ 
I believe we should be careful mocking or stubbing the class under tests. This may violate encapsulation and provide side effects in the future. To solve this, we can introduce another class that would work with the given arg results. Or we can call the command under test from the above using the `run(args)` method of the runner. 

This is an interesting discussion about this problem https://softwareengineering.stackexchange.com/questions/107361/is-it-okay-to-fake-part-of-the-class-under-test

I suppose in our case splitting an object would work perfectly.

metrics/cli/test/cli/command/update_command_test.dart
```diff
@@ -0,0 +1,189 @@
...
+        final option = argParser.options['config-file'];
+
+        expect(option.valueHelp, isNotEmpty);
+      },
+    );
+
+    test(
+      ".run() creates updater using the given updater factory",
+      () async {
+        whenCreateUpdater().thenReturn(updater);
```
solid-vladlaba, line 99: ✅ 
Maybe we should rethink this a little bit. I believe if our class has always the one and only one return object - preprogrammed - then we should create a stub implementation instead of mocking an object. For the class where we really need the mock implementation - we can create the separate Updater and use it to check with `.called(once)`.

solid-vladlaba (changes requested):
Well done, @solid-dimakoniaiev! I've added a couple of suggestions below - please let me know what you think!

solid-vladlaba (approved):
LGTM, thanks @solid-dimakoniaiev!:rocket:

